PR #43, opened 2022-03-20
Axis Aligned State Dragging

tdib:


frontend/src/hooks/useStateDragging.js
```diff
@@ -45,7 +59,7 @@ const useStateDragging = ({ containerRef }) => {
     const cb = e => {
       if (e.button === 0) {
         setDraggedState(null)
-        setDragStartPosition(null)
+        setDragOffset(null)
         e.preventDefault()
```
giraugh, line 63: Ideally we need to reset `dragCenter` as well

giraugh (approved):

PR #59, opened 2022-03-26
Implement Action History

giraugh:
Add functionality to project store for storing project history. In the future, after the project is edited by the user, call `commit` from the project store to add to the history.


frontend/src/stores/useProjectStore.js
```diff
@@ -1,6 +1,7 @@
 import create from 'zustand'
-import produce from 'immer'
+import produce, { current } from 'immer'
```
GRA0007, line 2: `current` not being used

GRA0007 (changes requested):

GRA0007 (approved):

PR #61, opened 2022-03-29
Swap from separate project models to singular project model

Prydeton:


giraugh (approved):
This code is top tier excellent stuff Max. Proud of you

PR #62, opened 2022-03-29
State Selection and Multi-state Dragging

giraugh:
Once merged, this PR will
- Implement state selection
- Implement a selection marquee box (that corrects for the view)
- Allow for dragging multiple states simultaneously if all selected
- Add nav dropdown items for 'Select All' and 'Clear Selection'
  -  The 'Clear Selection' item is currently implemented.
  - @GRA0007 do you have any advice / ideas for implementing Select All? I think we need to move the menu options from where they currently reside.

### DIFF
--- a/frontend/public/style.css
+++ b/frontend/public/style.css
@@ -5,6 +5,7 @@
   --primary-s: 84%;
   --primary-l: 52%;
   --primary: hsl(var(--primary-h) var(--primary-s) var(--primary-l));
+  --primary-highlight: #EB991D;
   --grid-dot: #ddd;
   --success: #23AF00;
   --error: #FC5353;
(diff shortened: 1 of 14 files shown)


frontend/public/style.css
```diff
@@ -5,6 +5,7 @@
   --primary-s: 84%;
   --primary-l: 52%;
   --primary: hsl(var(--primary-h) var(--primary-s) var(--primary-l));
+  --primary-highlight: #EB991D;
```
GRA0007, line 8: Hold on, shouldn't this colour just be the `--primary`? It's just used for the state stroke when selected right?

GRA0007 (changes requested):
We should definitely register our keyboard shortcuts somewhere other than the dropdown menus, and then just have a store we can pull the shortcut key symbols from for the dropdowns

frontend/public/style.css
giraugh: Haha yep, my bad. I kinda just assumed it was a different colour. I'll swap it over

GRA0007 (approved):

PR #63, opened 2022-03-30
User storage and authentication

Prydeton:
Implements user routes and Firebase authentication


backend/src/controllers/project.ts
```diff
@@ -1,31 +1,30 @@
...
-        message: error.message,
-        error
-      })
+  try {
+    const project = await Project.findById(pid).orFail()
+    return res.status(200).json({
+      project
     })
+  } catch (error) {
+    return res.status(500).json({
```
giraugh, line 15: WIll this respond with a `500` if the project doesn't exist? We might want to seperately handle that and return a 404 or remove the `orFail()` call and just return null

backend/src/controllers/user.ts
```diff
@@ -1,7 +1,44 @@
...
+  const { uid } = req.params
 
   // Retrieve user by id
+  try {
+    const user = await User.findById(uid).orFail()
+    return res.status(200).json({
+      user
+    })
+  } catch (error) {
+    return res.status(500).json({
```
giraugh, line 15: Same problem here, we prob don't want a 500 if the user doesn't exist. Just returning null is prob fine.

backend/src/routes/user.ts
```diff
@@ -0,0 +1,10 @@
+import express from 'express'
+
+import { createUser, getUser } from 'controllers/user'
+
+const app = express.Router()
+
+app.post('/', createUser)
+app.get('/:uid', getUser)
+
+export default app
```
giraugh, line 10: Ooh some fancy express router stuff, v nice :ok_hand:

backend/src/types.d.ts
```diff
@@ -16,4 +16,10 @@ export type meta = {
   dateEdited: string
   version: string
   automatariumVersion: string
+}
+
+declare module 'express-serve-static-core' {
+  interface Request {
+    uid?: string
+  }
```
giraugh, line 24: I've genuinely not known how to do this in the past XD - makes sense in retrospect

frontend/package.json
```diff
@@ -32,6 +35,7 @@
   "devDependencies": {
     "eslint": "^8.11.0",
     "eslint-plugin-react": "^7.29.4",
-    "parcel": "^2.3.2"
+    "parcel": "^2.3.2",
+    "process": "^0.11.10"
   }
```
giraugh, line 40: I don't think you need to have `process` as a dependency, its a built in node module. I might have missed something though

frontend/src/pages/Signup/Signup.js
```diff
@@ -0,0 +1,91 @@
...
+      setIsSubmitting(false)
+    }
+  }
+
+  return (
+    <form onSubmit={handleSubmit(onSubmit)} >
+      {error && (
+        <p>${error}</p>
+      )}
+      <p>email</p>
```
giraugh, line 74: Generally, `label` tags are preferred over `p` tags for labelling form inputs because you can associate them such that clicking on the label focuses the input. It's also better for accessibility reasons.

e.g
```html
<label htmlFor='email'>Email</label>
<input type='email' {...register('email')} />
```

giraugh (changes requested):
There's a couple small things to fix but overall is looking good :+1: Sorry it took ages to get around to this

Prydeton:
No longer being used as branch name changed to feat/backend-user-projects. Requested changes have been made.

Prydeton:
Closing as duplicate of #106

PR #64, opened 2022-04-01
JFLAP to Automatarium Conversion

tdib:
Given a `.jff` file, this package will convert it into Automatarium format


giraugh (approved):
LGTM :)

PR #68, opened 2022-04-02
State Creation

giraugh:
Once merged this PR will,
- Extract the view transform method from SelectionBox to be used generically
- Implement creationg states with the 'state' tool
- Split the transition tool into a 'transition' and 'state' tool
- Implement states rendering a default name using their id when one is not provided


GRA0007 (approved):
eeee

PR #69, opened 2022-04-07
Fix package json

giraugh:


GRA0007 (approved):
hm, I guess this is ok -_-

PR #90, opened 2022-04-27
Adjust grid snapping for groups of states

giraugh:
Fixes issue where states indivudally snap to the grid. New system involves an arbitrary state in a selection being deemed the "leader" with all other states following how that state snaps to the grid. I have an inkling that there may be a simpler or neater strategy if anyone wants to give it a shot. The relevant code is in the `useStateDragging` hook under the `GraphContent` component.


GRA0007 (approved):
tmlg

PR #99, opened 2022-04-28
State Actions

giraugh:
Once merge this PR will,
- Implement "Set as initial" context option
- Implement "Toggle is final" context option
- Create a system to allow actions to be dynamically disabled
- Change default project graph


GRA0007 (approved):
click click

Issue #109, opened 2022-05-06
Lambda as last transition


tdib:
When an automaton has a lambda as the final transition to an accepting state, the execution does not allow this and rejects the input

![image](https://user-images.githubusercontent.com/79500311/167110701-4be447dd-17d8-410d-a055-9b1bb275ff59.png)

PR #110, opened 2022-05-06
Fix FSA bugs

### LINKED ISSUE #107: Input consumed before reaching final state outputs undefined
Not sure of exact replication steps but it appears that rejected inputs (with the cause of rejection being not enough characters to consume rather than no possible transitions from the current state) prints undefined - this should probably just remove the "<read>:" part from the start of that input line

image

tdib:
Fixes the following:
- Undefined output when running out of input characters on rejected input
- FSA execution not attempting lambda as a final transition(s)
- Skip trace button (user had to press back a step to see accepted/rejected text)

All testing lab outputs *should* be correct and all cases handled:
- Accepted input
- Rejected input due to running out of characters
- Rejected input due to no possible transitions from given character
- Any lambda transitions :))))))))))))

Closes #107 #109 #111

### DIFF
--- a/frontend/src/hooks/useActions.js
+++ b/frontend/src/hooks/useActions.js
@@ -1,5 +1,5 @@
 import { useEffect, useMemo } from 'react'
-import { useProjectStore, useSelectionStore } from '/src/stores'
+import { useProjectStore, useSelectionStore, useViewStore } from '/src/stores'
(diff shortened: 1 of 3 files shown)


spacediscotqtt (approved):
LGTM

PR #123, opened 2022-06-05
Project Saving and Loading

giraugh:
## Completed
Once merged this PR will,
- Implement syncing projects list with backend
- Implement autosaving of projects
- Implement manual save action
- Setup a loading state for the editor page to ensure project is synced before being shown
- Create a spinner component
- Show a saving indicator on the menubar when there are unsaved changes (looking for feedback, is it annoying?)
- Randomly generate project names (This has been helpful for testing so leaving it in for now)
- Allow project names to be changed by clicking on project name (This is a temporary solution. Leaving final approach to @GRA0007)

## Overview of Saving Mechanism
There are two stores involved with saving, the `useProjectStore` and the `useProjectsStore`.

The `useProjectStore` stores the current state of the project. This is what is directly edited on the `/editor` page. This store is persisted to disk to persist the state of what project is "open".

The `useProjectsStore` stores all of the users projects regardless of whether they are open. For authenticated users, this store is a mirror of the projects stored on the backend. When you open any page after logging in, the list of projects are synchronized with the backends list of projects. This syncronisation process is handled by the `useSyncProjects` hook and re-occurs if the list of projects in `useProjectsStore` is updated.

When you navigate to `/editor`, a project must bet set in the `useProjectStore`. If not you are sent to `/new` where you can create a new project or open an existing one. If one is set, then the editor page displays a spinner while that specific project is synchronised. If the user is not authenticated (after authentication is done loading) then the loading state is disabled. Otherwise, the project with the matching `_id` is pulled from the backend and the dates are compared. If the backend project is newer then the project in the `useProjectStore` is overriden. After that process is resolved, the loading state is disabled and the editor page is displayed.
(This process happens in the `useSyncCurrentProject` hook)

While using the editor page, a periodic check is performed for whether an autosave needs to happen. If the state of the `useCurrentProject` project has changed since the last save then a save occurs. The saving process updates the project in the `useProjectsStore` store from the one in the `useProjectStore` store. This is then synced to the backend via the `useSyncCurrentProject` hook as mentioned previously. This save process can also be triggered manually using the `SAVE` action (triggered with Command+S).

As a final detail, the useProjectStore will update a `lastChangedDate` property whenever the project is updated and the saving process will update the `lastSavedDate` in the same store. The comparison of these dates is how the `saving...` status is displayed on the menubar and how the auto save process knows whether saving is necessary.

## To Do
- Investigate whether the changed date in `useProjectStore` can be automatically set using zustand middleware


tdib (approved):
> Show a saving indicator on the menubar when there are unsaved changes (looking for feedback, is it annoying?)

Dib loves a bit of feedback, it's perfect

> Randomly generate project names (This has been helpful for testing so leaving it in for now)

I think this is nice and will stop users from potentially having 100 "Untitled Project", I say we keep it but perhaps dim the name to better indicate that they can change it - This becomes a little tricky when they actually want the name like that (i.e. how to determine whether that's the permanent or temp name) but yeah. I also think the autosave feature should be disabled until a user actually makes a change to a file so it doesn't save a blank file every time they open one but not sure how easy that is.

Rest of the stuff seems to work but we definitely need a way to delete projects so they don't just pile up as empty projects (should be easy seeing as though we have `projectsStore` but could be wrong). Amazing work!

tdib:
Further note: If a change is made and the file is immediately left, the save isn't actually made

PR #124, opened 2022-06-06
Refactor Simulation Package

giraugh:
Oncer merged this PR will,
- Swap to typescript
- Reorganise tests
- Add more tests
- Setup read parsing (including range expansion) (w/ tests)
- Improve efficiency of `simulateFSA` method by using lambdaClosures instead of a lambda limit


giraugh:
@tdib we can fix this at the source of the problem

Prydeton (approved):
LGTM - Dib

PR #139, opened 2022-06-10
Finesse Testing Lab

giraugh:


tdib:
Automaton with initial and final state on same state causes output box to say accepted but with "error arrow"
image

tdib (approved):

GRA0007 (approved):

PR #143, opened 2022-06-10
Safari doesn't support user-select

GRA0007:
Just uses `-webkit-user-select` as well on the svg


tdib (approved):
Wow. What a comprehensive piece of work. This truly brought a tear to my eye. The way these characters have been types really inspires me to do more with my life. Thank you. LGTM

PR #145, opened 2022-06-10
You can now click to edit the project name in place

GRA0007:
also shh check out the saving indicator now


tdib (approved):
Very hot, nice work, amazing, LGTM

PR #147, opened 2022-06-10
Feat/home page import

tdib:
- Slightly reformats the new file page
- Allows importing of `.jff` (JFLAP) and `.json` (Automatarium) files
- Other file types will give the user a warning about the file types


GRA0007 (approved):
ldtm

PR #150, opened 2022-06-10
Update project model to store IDs as number

tdib:
Some issues were being caused as a result of some properties within the model being stored as string (e.g. initial state) and thus could not load properly


GRA0007 (approved):
ldtm

PR #151, opened 2022-06-10
Preferences

GRA0007:
Take that @Prydeton


tdib (commented):
Functionality-wise everything seems to be working but design-wise I think we need to have a team meeting - some of these colours are disgraceful to my eyes and you should apologise

I think purple saturation should go from 70 to 60, teal saturation from 100 to 80, and indigo be called indigo blue or blue :)

But more importantly (and no h8, but) I really don't vibe with the dark designs - I'm thinking it would be better to take a heavily desaturated version of the primary colour and apply it to both the fills and edges, and when a state/transition is selected it can be saturated and slightly darkened. I've whipped up a few examples of what this looks like and I personally prefer this to the white borders and very dark colours currently. Benji and I have discussed this and he doesn't like the coloured states so I'd like to discuss with the group and see what everyone thinks.
image
image
image

giraugh (approved):
LGTM

PR #155, opened 2022-06-11
Testing lab QOL

GRA0007:
- Allow tables to scroll horizontally
- Paste multiline text into the testing lab
- Use enter and delete to quickly make more rows
- Disable step buttons if there's no initial state


tdib (approved):
Very very nice work, love it, LGTM

PR #160, opened 2022-06-11
Keyboard shortcut guide

GRA0007:
image


tdib (approved):
- Being able to press on the buttons VERY nice touch
- My main issue is that I don't think that "move between multi-run cells" should include left/right arrows (like that's a given since it's text), and moving *between* cells is done with up/down.
- The whole menu feels a little long but it looks nicer than it probs would with multi column.
- Perhaps the top two options in the menu could do with a heading

I'm kinda strong on changing all arrow keys to just up and down but I leave the rest of these comments to your discretion.
LGTM

tdib:
Actually should fix this tsk tsk tsk
image

GRA0007:
@tdib lol my bad that was a mistake dw I'll fix the arrows

PR #162, opened 2022-06-12
Add project delete method and route

Prydeton:


tdib (approved):
LGTM amazing back work

PR #163, opened 2022-06-12
Fix JFLAP Importing

tdib:
For the most part appears to be broken
- Imported .jff files with one state, transition or comment would cause an issue because they would be represented as [undefined] and then try to be mapped: this should no longer put undefined in an array
- When a .jff file was imported while inside a file it would override the existing project in the store (including necessary metadata), preventing the ability to open testing lab and saving the file: fixed by first creating a new project and then spreading the imported file properties -- **I feel like this isn't the way to go but it allows for importing**

~~Side note: importing a JFLAP project of a different type (e.g. TM) gives an alert window of a message determined by us (first image below), but the console errors something a little clearer (second image). Wondering if it would be worth replacing the alert content with the thrown error or could there be some wacky errors?~~

### DIFF
--- a/frontend/src/hooks/useActions.js
+++ b/frontend/src/hooks/useActions.js
@@ -6,6 +6,7 @@ import { VIEW_MOVE_STEP, SCROLL_MAX, SCROLL_MIN } from '/src/config/interactions
 import { convertJFLAPXML } from '@automatarium/jflap-translator'
 import { haveInputFocused } from '/src/util/actions'
 import { dispatchCustomEvent } from '/src/util/events'
+import { createNewProject } from '../stores/useProjectStore'
 
 const isWindows = navigator.platform.match(/Win/)
 const formatHotkey = ({ key, meta, alt, shift, showCtrl = isWindows }) => [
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useActions.js
```diff
@@ -6,6 +6,7 @@ import { VIEW_MOVE_STEP, SCROLL_MAX, SCROLL_MIN } from '/src/config/interactions
 import { convertJFLAPXML } from '@automatarium/jflap-translator'
 import { haveInputFocused } from '/src/util/actions'
 import { dispatchCustomEvent } from '/src/util/events'
+import { createNewProject } from '../stores/useProjectStore'
```
giraugh, line 9: Prob use absolute import 😎

frontend/src/hooks/useActions.js
```diff
@@ -378,12 +379,22 @@ const promptLoadFile = (parse, onData, errorMessage='Failed to parse file') => {
...
+        onData({
+          ...project,
+          meta: {
+            ...project.meta,
+            name: input.files[0]?.name.split('.').slice(0, -1).join('.')
+          }
+        })
       } catch (error) {
-        window.alert(errorMessage)
+        window.alert(errorMessage + '\n' + error)
```
giraugh, line 397: I reckon a good use for string interpolation 🎈

giraugh (approved):
Looking good 👍 
(Just noted a few style changes)

PR #166, opened 2022-06-12
Power up transition input

### LINKED ISSUE #148: Transition input clickable enter
Do this
image

GRA0007:
Closes #148 

image

<3


tdib (approved):
I think optimally comments should use the same or a similar modal to the transitions but this LVGTM nice work!

PR #176, opened 2022-06-12
Pls

tdib:
f for fit
more reachable
PLEASE :(


GRA0007:
Change it in the hotkey panel as well and you have yourself a deal (yes I know it has to be updated in 2 places ugh we'll fix it maybe)

tdib:
Easy, done - the hotkeys menu was actually still set as shift + 1 so good thing this was noticed

GRA0007 (approved):
I removed the hotkeys file since we weren't using it for anything

PR #177, opened 2022-06-12
Use project name as downloaded file name

tdib:
Instead of prompting the user for a name when saving an automaton, it automatically names it based on the project name.

This is far more streamlined because they 1) don't have to type anything, 2) likely already have the project named what they'd want the file to be called.

In the worst case they can rename the file once it's downloaded but I believe the majority of the time they'll have it named properly.

Like if you agree

### DIFF
--- a/frontend/src/hooks/useActions.js
+++ b/frontend/src/hooks/useActions.js
@@ -70,18 +70,15 @@ const useActions = (registerHotkeys=false) => {
     SAVE_FILE_AS: {
       hotkey: { key: 's', shift: true, meta: true },
       handler: () => {
-        const fileName = window.prompt('What would you like to name this automaton?') // TODO: better prompt
-        if (fileName) {
-          // Pull project state
-          const { project } = useProjectStore.getState()
+        // Pull project state
+        const { project } = useProjectStore.getState()
 
-          // Create a download link and use it
-          const a = document.createElement('a')
-          const file = new Blob([JSON.stringify(project, null, 2)], {type: 'application/json'})
-          a.href = URL.createObjectURL(file)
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useActions.js
```diff
@@ -70,18 +70,15 @@ const useActions = (registerHotkeys=false) => {
...
-          const file = new Blob([JSON.stringify(project, null, 2)], {type: 'application/json'})
-          a.href = URL.createObjectURL(file)
-          a.download = fileName // TODO: prompt file location - might not be possible?
-          a.click()
-        }
+        // Create a download link and use it
+        const a = document.createElement('a')
+        const file = new Blob([JSON.stringify(project, null, 2)], {type: 'application/json'})
+        a.href = URL.createObjectURL(file)
+        a.download = project.meta.name // TODO: prompt file location - might not be possible?
```
GRA0007, line 80: Can you take a look at what I do to the file name when saving an image (useImageExport.js) to remove spaces and reserved characters

GRA0007 (changes requested):

frontend/src/hooks/useActions.js
tdib: Oh hm didn't consider reserved characters but why spaces?

GRA0007: Good question, I guess spaces are ok. If you're happy to keep spaces then please edit the image export code to keep them too

tdib: Coolio done :)

GRA0007 (approved):
Nice nice

PR #180, opened 2022-06-13
Prevent cancel from nulling comment value

### LINKED ISSUE #179: Empty comments can still be created
Cancelling the prompt still causes the state to be created (with null content)
image
image

GRA0007:
Fixes #179


tdib (approved):
LGTM

PR #181, opened 2022-06-13
Remove duplicates from transition reads

### LINKED ISSUE #170: Transition text still allows duplicate characters
Transitions can contain text such as "woo", but should instead strip the duplicate characters and be "wo"

GRA0007:
Fixes #170


tdib (approved):
LGTM

PR #187, opened 2022-06-13
State labels and names

GRA0007:
image


tdib (changes requested):
State names should be reflected in trace output
(also names that are too long go off the state - not sure if this can or should be truncated)
image
image

tdib (approved):
LGTM

PR #188, opened 2022-06-13
Implement trace tape functionality

giraugh:
image

### DIFF
--- a/frontend/public/style.css
+++ b/frontend/public/style.css
@@ -3,6 +3,7 @@
 
   --white: #FAFAFA;
   --black: #111;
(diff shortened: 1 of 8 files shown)


tdib:
The min-width on the trace tape is too small to show a single character
image
The tape should disappear with no input
image
Zooming in very close causes the trace tape to look tiny relative to the state - perhaps a min size could be set to prevent this?
image
Actual method to enable it needs at least a *little* pizzazz, an icon, description, something like that :)

tdib (approved):
LGTM except for the pointer >:((((

PR #200, opened 2022-06-14
Fix fit to zoom equation

### LINKED ISSUE #183: Fit to zoom strange behaviour
Fitting to zoom with an automaton arranged in a line fits towards the bottom of the screen - not a huge deal but worth noting
image

GRA0007:
Fixes #183 

can't stop won't stop


tdib (changes requested):
ngl I think a tiny bit more padding would be nice, I think it's currently 20, thoughts on 30?

Issue #164, opened 2022-06-12
Comments only showing two borders


tdib:
I'm not exactly sure how to replicate this but there appears to be some bugs with saving - this might be a result of corrupted projects but here's a screenshot
![image](https://user-images.githubusercontent.com/79500311/173226173-815d618a-ddc1-45fa-b7da-913b21f86ae8.png)

tdib:
I think I might actually know the issue - sometimes when projects save, they go to the bottom of the "your projects" list (even though it's the most recent), and if you click on a project from there something happens to do this but still not entirely sure why or what is happening.

GRA0007:
Closed as this is a browser bug

PR #201, opened 2022-06-14
Fix/comment text

### LINKED ISSUE #190: Dark mode comment text colour
Text colour on dark mode should be `var(--white)` instead of `var(--stroke)` - looks much nicer imo
image

GRA0007:
Fixes #190

So turns out there _was_ a problem causing #164, the comment bounding box didn't take the zoom into account when calculating its size, so if you edited a comment or opened a project at a different zoom, they would be clipped or have a larger bounding box.

I've fixed this by taking the zoom into account when setting the comment bounding box.

This PR also makes the comment text white in dark mode.


GRA0007:
@tdib 
Yeah I can't fix that, that one is definitely a browser issue :(

PR #202, opened 2022-06-14
Export image dialog

GRA0007:
Adds a dialog for exporting images with a nice preview. There are several options to choose from. Note that there is room to add a "dot grid" background option, but this will need to be recreated with svg for exporting so I'm not going to touch it now.

image

There's also an updated export menu that adds a shortcut for quickly copying to your clipboard.
image

@tdib to test if <kbd>Ctrl</kbd> + <kbd>e</kbd> works on windows


tdib (changes requested):
This one is amazing, I think this should say "quick copy to clipboard" to match up with the other ones cos it kinda feels out of place
Ctrl + e works perfectly
Also think we should add a little disclaimer on svg about how it won't copy to clipboard properly instead of having them thing it'd work
image

GRA0007:
I'm not too worried about the SVGs, fwiw exported SVGs also don't import correctly into illustrator etc. This is a separate issue we should look into but I don't think it's a priority atm.

GRA0007:
image
Ready to review @tdib

PR #204, opened 2022-06-15
Homepage spruce

GRA0007:
- Adjusted the homepage copy
- The FSA example image now adjusts to match your colour theme
- Real live working trace stepper
- Fixed a bug in the actual trace output where the state name wasn't used for the final line


tdib (approved):
Awesome work, LVVGTM

PR #208, opened 2022-06-16
Allow panning view with touchpad

GRA0007:
Mouse users must now hold ctrl to zoom with the scroll wheel. This also fixes the position slipping while zooming occasionally, an issue that was caused due to the position and zoom being set separately while zooming.

### DIFF
--- a/frontend/src/components/CommentRect/CommentRect.js
+++ b/frontend/src/components/CommentRect/CommentRect.js
@@ -20,7 +20,7 @@ const CommentRect = ({ id, x, y, text }) => {
         width: bounds.width*scale,
       })
     }
-  }, [containerRef?.current, text, scale])
+  }, [containerRef?.current, text, scale, x, y])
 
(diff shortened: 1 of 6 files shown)


GRA0007:
(also I checked mathematically and the zoom indicator on the bottom of the toolbar is accurate)

tdib (changes requested):
So it works on Windows fine (and presumably better on MacOS), but I do think the default for Windows should be to scroll by default rather than holding ctrl seeing as this isn't really a "scroll context" and as always a preference for this would be nice.

GRA0007:
> So it works on Windows fine (and presumably better on MacOS), but I do think the default for Windows should be to scroll by default rather than holding ctrl seeing as this isn't really a "scroll context" and as always a preference for this would be nice.

Ok as discussed this should be a separate issue

tdib (approved):
LGTM

PR #209, opened 2022-06-17
File thumbnails

### LINKED ISSUE #178: Recently edited projects being positioned on wrong side of project list
When a  project is edited and the user navigates back to the new project page, the most recently edited (or multiple) are positioned at the bottom instead of the top - a refresh fixes this but still worth noting

### LINKED ISSUE #206: File thumbnails

GRA0007:
Closes #206 

image
bam

Also fixes #178, turns out all of the dates were being saved as iso strings instead of milliseconds. You'll need to reopen all your projects to reset their save dates.


tdib (changes requested):
INCREDIBLE WORK I LOVE IT
I will say however (sorry) that the light background doesn't look too good on dark theme - I think it would look better if it matched up to your theme like you can with the export and the dark mode switch
image
Also there seems to be some weird alignment happening with files with long names and it looks like it might've been introduced in this branch so I'll chuck that in issues

GRA0007:
I can fix the text alignment in this branch, but they way thumbnails work is that they're generated when the file is opened, so I can't make them match your system theme, otherwise they would potentially be different if you change between dark/light frequently

tdib (approved):
Noice work benjer LGTM

PR #210, opened 2022-06-17
Use SFW adjectives and planets/moons for nouns

GRA0007:
no more uncooked fence sorry ewan


tdib (changes requested):
IMO the nouns are too sciency and difficult to pronounce - I think we should either revert back to a cleaned version of what we had or do something like a garden theme (e.g. spade, daisy, sunflower, fountain, grass, etc.)

tdib (approved):
Looks great to me after changing half of it 👍

PR #215, opened 2022-06-25
Fix saving issues

### LINKED ISSUE #211: Project isn't saved when changing the trace input or file options
Perhaps they are not `commit`ted? I'm not sure what triggers a save

GRA0007:
Fixes #211

Also changes the dates on the backend to be numbers, pending access from @Prydeton to run a migration to fix existing data.


giraugh (approved):
LGTM (even if filled with other random changes SMH)

PR #216, opened 2022-06-25
Add preference for zooming with ctrl key

### LINKED ISSUE #213: Ctrl zoom preference
On Windows the default zoom behaviour should be without holding ctrl, and on MacOS the default should be with (to allow for better trackpad control) - these should be able to change through preferences.

GRA0007:
Closes #213

### DIFF
--- a/frontend/src/stores/usePreferencesStore.js
+++ b/frontend/src/stores/usePreferencesStore.js
@@ -6,6 +6,7 @@ const usePreferencesStore = create(persist(set => ({
     theme: 'system',
     color: 'match',
     showGrid: true,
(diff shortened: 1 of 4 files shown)


frontend/src/stores/usePreferencesStore.js
```diff
@@ -6,6 +6,7 @@ const usePreferencesStore = create(persist(set => ({
     theme: 'system',
     color: 'match',
     showGrid: true,
+    ctrlZoom: !navigator.platform?.match(/Win/),
```
giraugh, line 9: Might be good to comment this line.

i.e literally this:
image

giraugh (approved):
LGTM

PR #217, opened 2022-06-25
Move preferences panel

### LINKED ISSUE #205: Preferences panel outside of project
If possible I think the ctrl+, shortcut should work outside the project - it's minor but there have been a few times I've tried to do it and it doesn't work so yee :)

GRA0007:
Closes #205

Moves the preference panel to the index file so it can be opened from anywhere on the site. Also added a preferences button to the new file screen.


giraugh (approved):
LGTM B)

PR #218, opened 2022-06-25
Prevent user from navigating away if project unsaved

### LINKED ISSUE #118: Before unload warning in editor if changes unsaved
https://stackoverflow.com/questions/71572678/react-router-v-6-useprompt-typescript/71587163#71587163

GRA0007:
Closes #118

If the user clicks the Automatarium logo and the project has unsaved changes, it will instead save and then navigate.


giraugh (approved):
Looks good to me. I reckon it would be good to eventually centralise / de-dupe the saving/syncing logic. Potentially we could add a save action to the projects store.

PR #219, opened 2022-06-25
Remove plurals from new project names

GRA0007:


giraugh (approved):
Looks cooked (in the best way) and also very GTM

PR #220, opened 2022-06-25
Use input dialog for all inputs

GRA0007:
No more window.prompt in the project 🎉 

Comments, state labels and state names all now use input dialogs.


giraugh (approved):
Looks good. Might be good if the custom events followed the `ctx:type` scheme. The comment icon might also look a lil bit like a button

PR #222, opened 2022-06-25
Use the dev firebase in prod

GRA0007:
In the future we can switch to a different account, potentially managed by RMIT


giraugh (approved):
LGTM for now

PR #224, opened 2022-06-25
Removed unneeded UUID package

GRA0007:
This functionality is now supported in all browsers natively


tdib (approved):
LGTM
I have mentioned doing UUIDs like this before :pensive:

PR #225, opened 2022-06-25
Fixed dark grid incorrectly used on project cards

GRA0007:


tdib (approved):
LGTM - I was about to tell you about this but you're too faste 😈

PR #227, opened 2022-06-25
Fixed arrowheads being black when exporting an image in dark mode

GRA0007:
This occurred if the arrowhead was selected before exporting


giraugh (approved):
Noice LGTM

PR #235, opened 2022-06-26
Multi-run lambda placeholder

GRA0007:
Uses lambda symbol for multi-run text input placeholder to indicate that it will be run with an empty input.

image

This PR also prevents all of the multi-run inputs from being removed, keeping at least one.


giraugh (approved):
Hype LGTM :ok_hand:

PR #236, opened 2022-06-26
Refactor padding to margin

GRA0007:
Image exporting options now use the word "margin" instead of padding.

This PR also constrains the margin between `0` and `500` to prevent larger images from being generated (which may crash the system).


giraugh (approved):
LGTM

PR #244, opened 2022-07-12
Info panel state names fix

### LINKED ISSUE #243: State transition function and table don't reflect state names
State transition function and table only show the state prepended with prefix (e.g. q12) rather than the state name. As shown below, the state name is "error" (original state name was q12), but it is still showing as q12 in the "b" column and second row respectively.
image
image
image

tdib:
- The transition function and table have been swapped - the function increases in size a lot longer and I've found myself most frequently accessing the table
- The state names are now shown instead of the ids (to reflect any renamed states)

An image of the updates are shown below:

Closes #243

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/Info/Info.js
+++ b/frontend/src/components/Sidepanel/Panels/Info/Info.js
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useMemo, useCallback } from 'react'
 import { validTransitions, resolveGraph, closureWithPredicate } from '@automatarium/simulation'
 
 import { useProjectStore } from '/src/stores'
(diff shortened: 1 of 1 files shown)


frontend/src/components/Sidepanel/Panels/Info/Info.js
```diff
@@ -47,6 +53,7 @@ const Info = () => {
 
     return map
   }, [states, alphabet, resolvedGraph])
+  console.log(transitionMap);
```
GRA0007, line 56: naughty naughty

GRA0007 (approved):
Was never marked as a reviewer 🥲

tdib:
You were pretty much there when I wrote it so I thought whoever saw it could review 🤷

frontend/src/components/Sidepanel/Panels/Info/Info.js
tdib: what do you mean by this 👀 nothing happened, nothing at all.

PR #250, opened 2022-07-20
Use screenToViewSpace and viewToScreenSpace when showing dialogs

### LINKED ISSUE #241: State label and name prompt appearing in wrong location
Trying to change q15 opens the prompt in the following location
image

There seems to be something going on with the calculation of positioning certain things - the same thing happens with comments, being placed very far from the cursor position

GRA0007:
Fixes #241


frontend/src/components/InputDialogs/InputDialogs.js
```diff
@@ -70,7 +71,9 @@ const InputDialogs = () => {
   const saveComment = () => {
     if (value && !/^\s*$/.test(value)) {
       if (dialog.selectedComment === undefined) {
-        useProjectStore.getState().createComment({ x: dialog.x, y: dialog.y, text: value.trim() })
+        const pos = screenToViewSpace(dialog.x, dialog.y)
+        console.log(pos)
```
tdib, line 75: what. is. this?. benjer?.?>!

PR #251, opened 2022-07-20
Check if left click is pressed to show ghost state

### LINKED ISSUE #248: Right click shows ghost state
When the state tool is selected and the user right clicks, the ghost state is shown. It does not place, and right click functionality works as expected but this ghost state should not appear on right click.

GRA0007:
Fixes #248


tdib (approved):
Very nice LGTM

PR #253, opened 2022-07-20
Extract _id and userid from exported JSON

### LINKED ISSUE #249: Exporting Automatarium file keeps project ID
When exporting Automatarium (.json) files, the `_id` attribute is exported along with the file, which causes some issues when importing. For example, attempting to export a file and then re-import it (acting as a duplicate function) causes no new file to be created, but a single project to be overridden.
There may be other attributes unwanted in the file export.

GRA0007:
Fixes #249


tdib (approved):
LGTM

PR #265, opened 2023-03-06
Update with files from last semester

ire4ever1190:
Adds the last group's code into the repo

I got a [temporary deployment](https://admirable-salmiakki-2c1be2.netlify.app) if anyone is curious what the site is now like (No backend running currently, but seems no changes were made to it anyways)

From looking at the documents we were given the new features are
- Rewritten execution algorithm to use BFS instead of recursively solving
- Supports PDAs (Closes #53)
- Supports Turing machines (With tape lab to provide UI for stepping through execution)
- Deleting states and transitions

PR #289, opened 2023-03-16
Automatarium V2

ire4ever1190:
Merges the work from last semesters into main since them finishing is basically a new release.

This contains all their commits from #265 excluding my commit (Since that wasn't part of their work)


ire4ever1190:
Think going forward we should try and merge into main after every sprint

LvandoApps (approved):
looks good

PR #290, opened 2023-03-20
Save the project to the store when renamed

### LINKED ISSUE #286: Project renaming does not save if project is exited quickly
When the name is changed while working on a project, the name persists in the window but if you navigate to the home screen too quickly the new name is not saved.

ire4ever1190:
Closes #286 

This is done instead of changing the `lastChangeDate` so that the change is done immediately instead of waiting for the next save cycle or user navigating away


LvandoApps (approved):
beautiful work

PR #291, opened 2023-03-20
Select the right simulation function when running multi trace

### LINKED ISSUE #273: Multirun doesn't select the right function for simulating the project
[Reproducible example](https://pastebin.com/raw/EZVSnCvz)

In the multirun for the PDA add strings (should be there if you load the JSON file)
- `AAA` (Should fail)
- `AAAA` (Should pass)

Building the same machine in JFLAP gives those results but Automatarium says they both pass.
Also weird, `AAAA` doesn't pass in normal run since it jumps to the final node too fast (this is a separate issue it seems)

More investigation finds that this line is the problem
https://github.com/automatarium/automatarium/blob/56feb729b3b8a33f6d48101cc0480f8849322dc9/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.js#L355
Since it means it always simulates multirun as an FSA (hence it ignores the stack requirement).

This logic error seems to appear in a few places. We should probably make a function `simulateGraph` that properly chooses how to simulate the project instead of all the `if`'s that are currently being used

ire4ever1190:
Closes #273 

The issue before was that in some places the `if` statements didn't properly switch between FSAs, PDAs. and TMs and some places didn't have any branching and just also simulated FSAs

This PR adds `runSimulation` which runs a provided input with the correct simulation function depending on the `projectType`

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.js
+++ b/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.js
@@ -50,15 +50,19 @@ const TestingLab = () => {
   const setPDAVisualiser = usePDAVisualiserStore(state => state.setStack)
   // const stackInfo = usePDAVisualiserStore(s=>s.stack)
 
-  // Execute graph
-  const simulateGraph = useCallback(() => {
+  /**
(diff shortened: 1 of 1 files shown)


LvandoApps (approved):
lovely

PR #294, opened 2023-03-20
Use `actions/setup-node` to cache yarn dependencies

ire4ever1190:
This should reduce the time that CI takes (~1.5 minutes is spent just getting dependencies)


SidhraFernando-Plant (approved):
LGTM - will be great to have this working much faster 😊🎉

PR #296, opened 2023-03-23
Fix default saving and formatting in TM

### LINKED ISSUE #282: Enter doesn't confirm transition for TM
Pressing enter doesn't confirm a TM transition, no matter which field is selected. Pressing enter on any field should confirm the transition, although a default value for direction will need to be set (R makes the most sense imo) as direction cannot be blank. If not a default will have to check that direction is not blank.

LvandoApps:
Closes #282

This PR will fix the default saving problem that occurred in the turing machine templates, this way you can press enter instead of a button in order to save AND you can save a default template of "lambda;lambda, R" (R felt most natural as that is the usual direction of a tape), as well as fix the formatting such that instead of it being horizontal, it is vertical just like the rest of the automata.

### DIFF
--- a/frontend/src/components/InputDialogs/InputDialogs.js
+++ b/frontend/src/components/InputDialogs/InputDialogs.js
@@ -216,9 +216,9 @@ const InputDialogs = () => {
             value={read}
             onChange={handleReadIn}
             onKeyUp={e => e.key === 'Enter' && save()}
-            placeholder={'λ'}
+            placeholder={'λ\t(read)'}
             style={{
-              width: '8ch',
+              width: `calc(${dialog.type === 'comment' ? '20ch' : '12ch'} + 3.5em)`,
               margin: '0 .4em',
               paddingRight: '2.5em'
             }}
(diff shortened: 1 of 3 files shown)


ire4ever1190 (changes requested):
Looks good :ok_hand: 

Just accept that one change then it should be good to merge.

frontend/src/components/InputDialogs/InputDialogs.js
```diff
@@ -180,88 +179,84 @@ const InputDialogs = () => {
...
-    setRead(input[input.length - 1] ?? '')
+    const value = input[input.length - 1]
+    setRead(value)
   }
+
   function handleWriteIn (e) {
     const input = e.target.value.toString()
-    setWrite(input[input.length - 1] ?? '')
+    const value = input[input.length - 1]
+    setWrite(value)
```
ire4ever1190, line 189: Not to sure how I triggered it but got a null access error because of this change. Think using the null coalescing is best idea unless there was a reason you changed it

LvandoApps (commented):
not intentional, mustve forgotten to change it back while experimenting, feel free to change it back

ire4ever1190 (approved):
Looking great :+1: going to be much nicer UX when making transitions

Issue #301, opened 2023-03-25
Merge graph types together


ire4ever1190:
Currently there are multiple different graph, state, and transition types that get converted between each other but have mostly the same structure. Instead we should use the abstract graph type that is used for solving in all these cases.

This will remove the need for conversion and means we can put duplicate code as methods inside the class

Issue #288, opened 2023-03-14
Add tests for `simulation-v2`


ire4ever1190:
Currently there are no tests for `simulation-v2` which means performing any refactors or bug fixes could cause regressions. Having tests will also give a clearly defined behavior for how we expect the matching to happen

Will need to
- [x] Port tests from `simulation` for FSA's
- [x] Add new tests for expected behavior for PDA's and TM's
- [x] Add test running to the CI

Note: This won't fully stop regressions, but should least catch the major issues

PR #302, opened 2023-03-25
Refactor both simulation packages into one

ire4ever1190:
> I am become deleter, the destroyer of code

This refactor combines both `simulation` and `simulation-v2` into one package. It was useless having the two packages since the old package was only used for a few functions (Which have been moved into `simulation-v2`). This also renames `simulation-v2` into `simulation` to reflect the change (But the version number has been bumped to 2.0.0)

This also ports the tests from `simulation` to work with the new code (Part of #288). There are some weird types in this refactor which mostly stem from the issue discussed in #301. So quite a few functions specify they take an `FSA` graph when in reality then just take a normal graph. This reflects how the code is currently used, but will be cleaned up in a future PR

There might be some regressions due to the large changes but haven't found any during my testing. Will also be able to add tests for PDA's and TM's much more easily now with this refactor to minimize regressions in future


LvandoApps (approved):
looks good to me, no logic errors at least, how come all the snake-case to camel-case though?

ire4ever1190:
Oh the linter was complaining since the code style says camelCase for variables

PR #305, opened 2023-03-26
Add git conventions into contribution guide

ire4ever1190:
This puts the git information given to us by @tdib into the contribution guide so it can be easily found


LvandoApps (approved):
good idea, approved

PR #307, opened 2023-03-26
Remove `yarn-error.log`

ire4ever1190:
Removes a left over yarn error log file and adds them to the `.gitignore` so they don't accidentally get added again


tdib (approved):
Sounds like a great plan in fact, LGTM

Prydeton:
This is required as per Yarn-error log documentation. Please revert this change.

PR #313, opened 2023-03-30
PDA and TM comment modals fixed

LvandoApps:
This fixes an issue with the PDA comment modals such that they no longer display the incorrect number of input fields (used to display three, now displays only one). On top of this, TM comment modals were fixed such that it is now no longer just the transition inputs and is now a proper comment modal. File was also tidied up to make future viewing easier.

### DIFF
--- a/frontend/src/components/InputDialogs/InputDialogs.js
+++ b/frontend/src/components/InputDialogs/InputDialogs.js
@@ -278,8 +278,7 @@ const InputDialogs = () => {
       )}
     </Dropdown>
     )
-  }
- else {
+  } else {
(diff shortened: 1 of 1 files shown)


ire4ever1190 (changes requested):
Looking good :rocket: 

Could the code be cleaned up so it isn't spaghetti of if's? Like there is still quite a bit of ifs inside components but imo would be nicer to split the logic up to return a different thing depending on `dialog.type` (Would probably require PDA's to get its own type like TM's)

ire4ever1190 (approved):
It solves the problem :+1:

PR #314, opened 2023-03-31
Fix tracing tape for TMs

### LINKED ISSUE #270: Tracing tape in Turing machine is broken
To reproduce
- Go into Turing machine project
- Enter some input
- Select trace tape

Page will then go blank due to some undefined access issues

ire4ever1190:
Closes #270 

Adds some more checks (Was missing a check if remaining was undefined) and simplifies the logic in places (There was a lot of magical minus 1's in places) which fixes stepping through the stepping tape


LvandoApps (approved):
Certainly some interesting logic, tidied up nicely now though well done

PR #315, opened 2023-04-02
Add linter to `jflap-translator` package

ire4ever1190:
This makes the translator package get linted also.

This also cleans up the root `package.json` to remove a lot of repeated scripts by using the `workspaces` command to run linting on every workspace

PR #316, opened 2023-04-02
Support importing PDAs and TMs from JFLAP

### LINKED ISSUE #267: Support importing other JFLAP types
#265 adds support for PDA's and turing machines but the importer doesn't work for those machines

ire4ever1190:
Closes #267 

Rewrites the JFLAP importer to be in typescript and now supports importing both PDAs and TMs from JFLAP

This also adds unittests for both the existing test files and for some PDA and TM machines

This is currently a draft since I'm waiting for #315 to be merged first which this is based on top of


SidhraFernando-Plant (approved):
LGTM, nice job on the tests 💯

One thing that sticks out to me is that some of the typescript conversion here overlaps with what I've done in the refactor/jstots branch for useProjectStore, we've used a few of the same types (state, transition, etc). I stored these types in a specific type folder/file to be imported as needed as these types are used widely across the project, so perhaps convertJFLAP.ts could be modified later on to import from the type file. On this note, see comment re FrontendGraph type.

Another small thing is that it would be nice to have PDA and TM imports match their project colours (ie. red and purple respectively). Not that important and defs out of scope for this PR but we could open an issue for this in the future.

packages/jflap-translator/src/convertJFLAP.ts
```diff
@@ -0,0 +1,160 @@
...
+ * Comment that is placed somewhere on the project
+ */
+export type Comment = {
+  id: number
+  x: number
+  y: number
+  text: string
+}
+
+// TODO: Replace with whatever type the frontenders come up with for the typescript conversion
```
SidhraFernando-Plant, line 58: Note that this has been completed in the branch refactor/jstots with the Project type. Once feasible this type can be imported from types/ProjectTypes (new folder on refactor/jstots branch)

ire4ever1190: Nice yeah meant to get back and update that. Will rebase and then use the types from the refactor.

How many stores are left? Would be good to either finish them off or just merge in what we have

ire4ever1190:
Adding colours shouldn't be too difficult and seems in scope so will add that into this PR :+1:

packages/jflap-translator/src/convertJFLAP.ts
SidhraFernando-Plant: There's still 6 left but the remaining ones are significantly smaller than useProjectStore so will take less time. I think it could be worth merging in what we have so far though, because the project ones are pretty important and would be good to have them there without waiting for the other 6 (I should be able to get them done in the next few days I'm just a bit busy with other assignments atm). Should I open a PR to merge in what we have?

ire4ever1190: Yeah would be good to merge in what we have

Then I can refactor the weird types in the simulator also to use what the frontend uses

frontend/src/stores/useProjectStore.ts
```diff
@@ -24,7 +24,12 @@ import {
   DEFAULT_PROJECT_COLOR
 } from '../config/projects'
 
-export const createNewProject = (projectType: ProjectType = DEFAULT_PROJECT_TYPE): Project => ({
+/**
+ * A stored project has an extra `_id` field which is used to tell identify it
+ */
+type StoredProject = Project & {_id: string}
```
ire4ever1190, line 30: Projects used by the store require an id but don't require one outside. This was the name I came up with but open for suggestions of a better name

SidhraFernando-Plant: SGTM 👍🏻

PR #317, opened 2023-04-03
Update main with latest sprint

ire4ever1190:
Merges commits from the latest sprint back into main


LvandoApps (approved):
Looks good, all other issues we do from this point on until the end of sprint 2 will be apart of the next batch to be merged into main (to be done after end of sprint 2, like how this was done after sprint 1)

PR #321, opened 2023-04-04
Refactor `eslint` to use a shared config

ire4ever1190:
Was fixing issue with `backend` erroring during linting when using windows line endings (Since they were explicitly disallowed for some reason). While doing that I thought it would be good to use a shared config so that we can easily set these settings across the entire monorepo without needing to change each individual config.

While doing that I found `backend` wasn't getting linted properly so fixed that up which is why the backend has changes


ire4ever1190:
yarn didn't like the extra lockfiles that were there but now the frontend doesn't build without them :thinking: 
They shouldn't be there anyways since workspaces [should only have one lockfile](https://classic.yarnpkg.com/lang/en/docs/workspaces/) in the root

Will fix in another PR first since the changes are out of scope for this one

LvandoApps:
It would probably be best to fix it in this PR (we don't want failing builds in dev after all), or you could make a separate PR that fixes the issue, merge it into dev, and then update this feature branch with those changes (it's up to on whether you think the changes are large enough to warrant it's own feature branch).

Either way I'll accept this when it's able to be deployed in Vercel.

ire4ever1190:
Fixed in #322

LvandoApps:
My bad didn't look through the rest of the PR's before commenting haha, I'll accept them both now.

LvandoApps (approved):
Accepted

LvandoApps:
NOTE: This PR fixed was fixed by #322, as such deployment of #321 into dev despite having a build failure is fine given that the merging of #322 into dev is done immediately afterwards.

PR #322, opened 2023-04-04
Remove lockfiles from workspaces

ire4ever1190:
Small cleanup to remove the unused lockfiles.

[Parcel doesn't allow specifying the root](https://github.com/parcel-bundler/parcel/issues/7206) so I added a `.hg` file which it can use to find the root.

Future refactor would be switching to vite which at a glace seems to support specifying the root. Although that would be a larger undertaking and might interfere with the current typescript refactor so will leave that for future me problem


LvandoApps (approved):
Hopefully this will allow build to pass in dev

LvandoApps:
NOTE: This PR fixed #321, as such deployment of #321 into dev despite having a build failure is fine given that the merging of #322 into dev is done immediately afterwards.

PR #323, opened 2023-04-04
Begin migrating stores to TypeScript

SidhraFernando-Plant:
This PR begins migrating the stores files to TypeScript, with the following files being migrated:
- useExportStore.ts
- usePDAVisualiserStore.ts
- usePreferencesStore.ts
- useProjectsStore.ts
- useProjectStore.ts

A types folder has also been created to group types that are commonly used throughout the project for easy importing. 6 stores still need to be migrated to TypeScript.


ire4ever1190 (approved):
Very cool :rocket: going to make refactoring other stuff a lot easier.
I just cleaned up some of the types so that they weren't `any` or unrestricted

frontend/src/types/ProjectTypes.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Possible directions that a TM transitions can move
+ * - Left
+ * - Right
+ * - Stay
+ */
+export type TMDirection = 'L' | 'R' | 'S'
+
+// May be different types for each automata
+export interface AutomataTransition {
```
ire4ever1190, line 55: Fine for now, but think we should have different types like the other packages have.
e.g. have a base `Transition` type which has the common stuff and then form other types 
```ts
export interface PDATransition extends AutomataTransition {
  push: string
  pop: string
}
```

SidhraFernando-Plant: Yeah for sure 👌

PR #324, opened 2023-04-04
Refactor dialogs to typescript

### LINKED ISSUE #311: TM doesn't properly save default direction when making transition
### Description of the bug

The direction isn't saved to the project when the default direction is selected (i.e. If the user makes a transition and doesn't make a direction, the UI shows `R` but that isn't saved in the project)

### To reproduce

Make a TM that goes through a few states reading characters (don't specify the direction when making the transitions) and then try to run it. Something like this would work

image

### Expected behavior

Transition maker should properly save the transition if the user doesn't select one

### Additional information

The stepping lab for some reason works even though the simulation fails, I'm assuming the stepping lab properly passes of the graph with the direction filled in

Caused by #296

ire4ever1190:
Refactors the input modals to typescript. Was testing out the types in the project stores (Discovered a few problems which are fixed here) and also testing how we would start moving other files to typescript.

I think the compiler file is correct, but since parcel is having import problems it will say that the imports are wrong (Just ignore that for now, will work on soon). This causes problems with IDE not providing proper hinting so I definitely will fix soon

Important to note, make sure you rename `.js` files to `.tsx` if converting a typescript file that uses tsx features

Inadvertently closes #311 (Sorry for stealing the issue Lachie, but typescript was screaming at me)


SidhraFernando-Plant (commented):
I think it could be good to add some more explicit typing in the TypeScript conversion, as a lot of the file has any types so I feel we are missing some of the value of converting to TypeScript. The dialog types at the top of the file are nicely defined, perhaps we could use them a bit more throughout the file? Open to hearing thoughts on this

ire4ever1190:
They are used throughout the file but with type inference it isn't really noticed

For example at the bottom it in the switch statement if you try and access fields that aren't defined for that modal (It changes modal type in each branch depending on `dialog.type`) it throws an error

ire4ever1190:
Think the `handleStuff` functions do have implicit any types so will fix those up though

SidhraFernando-Plant:
I'm not sure if my VSCode is being funny, are the types of the useState hooks being inferred? VSCode is telling me that all of the values and setter functions for the useState hooks are any type but this doesn't seem right to me. One of my main concerns was all of the const functions defined at the start so if the hooks are implicitly typed that's fine. 

I thought VSCode shows implicit type but maybe it doesn't, but anyway bottom line I think it has caused me to think the typing was a lot looser than it actually is because i thought i was seeing the inferred types but don't think I was

ire4ever1190:
Have you ran `yarn install` in the root folder? I added the react types which you might be missing

But the types are inferred since we give them a default. The only one I needed to explicitly state was the TM direction one

SidhraFernando-Plant:
Yep okay hadn't run yarn install recently so now can see the inferred types on the hooks 😅 I have a few more things to add soon sorry I'm a bit slow with replies been a bit busy with other assignments but I haven't forgotten about this will update soon

SidhraFernando-Plant:
The only idea I can think of for typing `useEvent` is this:
- In useEvent.ts pass the first two arguments `name` and  `handler` as an object
- Make a type for this object something like this
```
type UseEventArgs = {
  name: 'editTransition',
  handler: { detail: { id: number } }
} | {
  name: 'editComment',
  handler: {detail: {id: number, x: number, y: number}}
}
```
In each tsx file that utilises `useEvent`, we could define the name/handler pairs for that file and export. Then all of the types can be imported into `useEvent.ts` and `UseEventArgs` can be defined as a union of the pairs from each file. Then the pairs don't all have to be defined in `useEvent.ts`.

The main problem is that I'm not sure how this would be made backwards compatible so it might be best to do it all later on as a bigger refactor. There is also probably a better way of doing this, this is all I could think of but I'm sure there's something else out there. Simply some thoughts on the matter

ire4ever1190:
Understandable, the assignment grind before the break is always a pain :skull: 

Yeah certainly will be a big refactor to do it. Would certainly be cool if it works well

SidhraFernando-Plant (approved):
Sorry for the delay, my VSCode is finally showing all of the types properly though so I can see they are all nice and well 😊 Thanks for your patience

PR #325, opened 2023-04-04
Cache eslint

ire4ever1190:
Turns on caching when running `eslint` to make it faster. Makes linting ~50% faster on my machine


LvandoApps (approved):
Personally didn't find the linting in the CI to be too much of an issue time wise, but always appreciate more efficiency where applicable, looks good

ire4ever1190:
Nah this has no affect on performance of the CI (since the cache isn't carried across runs)

It's more for local since it can be slow at times (and I have it has a `pre-commit` hook so always nice when it runs fast)

PR #326, opened 2023-04-05
Fix TS and JS language server not resolving imports properly

ire4ever1190:
Adds in config files to specify where the base folder is for root imports


SidhraFernando-Plant (approved):
Thanks for sorting this out ✅

PR #327, opened 2023-04-05
Fixes running into `config` being undefined when loading fresh site

ire4ever1190:
Discovered problem in #324 (If you try and view the deployment there you will see the problem)

The values need to be null or else it doesn't think they need to be initialised (i.e. it will save the values which means it can't access properties like config since its an empty object)


LvandoApps (approved):
Spooky

Issue #319, opened 2023-04-03
Template Creation for States/Transitions


LvandoApps:
### Proposed feature or solution

Copy and paste a section of states/transitions and be able to add it into a "Templates" section that is saved across all instances. At any time the user will be able to go into this "Templates" section and click on the template, then like a ghostly version of the template will appear and the user can move it around the automata screen, then when they left click again it places it. Numbers assigned to each state will be based on what states already exist, and the user is able to delete a template whenever they want.

SidhraFernando-Plant:
Closed with #381 and #348

PR #328, opened 2023-04-05
Refactor useSelectionStore to ts

SidhraFernando-Plant:
Refactors useSelecctionStore to TypeScript.

I will be using this store a lot for issue #319 so I wanted to get it done before I made a start on that issue.


frontend/src/stores/useSelectionStore.ts
```diff
@@ -17,17 +30,17 @@ const useSelectionStore = create(set => ({
...
       .project
       ?.transitions
-      ?.map(t => t.id) ?? [],
+      ?.map((t: AutomataTransition) => t.id) ?? [],
     selectedComments: useProjectStore
       .getState()
       .project
       ?.comments
-      ?.map(c => c.id) ?? []
+      ?.map((c: ProjectComment) => c.id) ?? []
```
ire4ever1190, line 43:
```suggestion
      ?.map(s => s.id) ?? [],
    selectedTransitions: useProjectStore
      .getState()
      .project
      ?.transitions
      ?.map(t => t.id) ?? [],
    selectedComments: useProjectStore
      .getState()
      .project
      ?.comments
      ?.map(c => c.id) ?? []
```
`s`, `t`, and `c` are inferred since `map` knows the type of the array its operating on

frontend/src/stores/useSelectionStore.ts
```diff
@@ -1,8 +1,21 @@
-import create from 'zustand'
-
+import create, { SetState } from 'zustand'
 import { useProjectStore } from '/src/stores'
+import { AutomataState, AutomataTransition, ProjectComment } from 'src/types/ProjectTypes'
+
+type SelectionStore = {
+  selectedStates: AutomataState[],
+  selectedTransitions: AutomataTransition[],
+  selectedComments: ProjectComment[],
```
ire4ever1190, line 8:
```suggestion
  selectedStates: number[],
  selectedTransitions: number[],
  selectedComments: number[],
```
These get mapped to ID's later which are numbers

ire4ever1190 (changes requested):
Sorry for the delay!

Looking good :+1: just a few changes

Always good to see another file converted to ts :rocket:

frontend/src/stores/useSelectionStore.ts
ire4ever1190: This change will cause errors in the other stuff, make sure to fix those also :+1:

frontend/src/stores/useSelectionStore.ts
SidhraFernando-Plant: Hmm okay IDE was telling me they were any so I thought it was a weird case where they weren't being inferred but I now realise that that's because of the issues with the imports/hinting. I realise at this point i should really just not trust the IDE lol but will double check with the imports next time

SidhraFernando-Plant:
No worries, still a very quick turnaround 🤠

frontend/src/stores/useSelectionStore.ts
SidhraFernando-Plant: After pulling I have gotten the fix for the import errors yay 🎉 So shouldn't be getting misled by that again (will also just distrust IDE from now on ofc)

frontend/src/stores/useSelectionStore.ts
SidhraFernando-Plant: Good pickup will shuffle it around

ire4ever1190 (approved):
One step closer to fully typed frontend :rocket:

PR #330, opened 2023-04-06
Check types in CI

ire4ever1190:
This runs the typescript compiler to check for any type mismatches.

Had to fix up some files to make the types be correct


LvandoApps (approved):
Looks good

PR #331, opened 2023-04-06
Fix default transition value in TM

LvandoApps:
This PR just ensures that the placeholder for transition is present in the turing machine rather than it just being 'R', which doesn't look very nice. The saving of default value is still kept in, this is just a visual bug fix more or less.


ire4ever1190 (approved):
Very nice :+1: 
thanks for fixing my oopsies

PR #332, opened 2023-04-09
Support reordering graph states

### LINKED ISSUE #234: Reorder state IDs
Option to reorder state IDs based on the transition flow.

ire4ever1190:
Closes #234 

Adds ability to reorder states in the graph (Command is done via right click menu). This is done using a flood fill that goes breath first when there is multiple options but continues along states that have single transitions. This is done so that continuous transitions come one after another. The reason a topological sort wasn't chosen was because that wouldn't work with cycles

**Before**
image

**After** 
image

(Notice how the left side is in order)

The ordering isn't perfect in some cases (especially when the graph is spaghetti) but for most cases it works well.


giraugh:
This looks so hype

tdib:
Yeah this is spicy hot

GRA0007:
Love this

LvandoApps (approved):
Looks sweet as

PR #333, opened 2023-04-09
Don't save blank projects

### LINKED ISSUE #226: Don't save blank projects
Disable auto saving until the user does something with the project (i.e. don't save as soon as it's opened)

ire4ever1190:
Closes #226 

This checks  $comments + states + transitions \geq 0$ before saving so that we don't save projects with nothing in them.

This still saves empty projects if you rename them since if you are giving it a name then you probably want it to stick around


LvandoApps (approved):
Good stuff

PR #334, opened 2023-04-09
Fix image export on firefox

ire4ever1190:
Firefox requires SVG elements to have a defined width and height if we want to draw them to a canvas

See bug: https://bugzilla.mozilla.org/show_bug.cgi?id=700533


SidhraFernando-Plant (approved):
Nice job finding this one and sorting it out

PR #335, opened 2023-04-09
Centre graph when loading project

### LINKED ISSUE #258: Center graph on project load
When a project is loaded, the view is centered at 0,0, and a graph may be positioned elsewhere, appearing as though there is nothing on the project.

chrome_OXZMT7LJm7

Preferably, the solution shouldn't simply fit the project to view on load (i.e. perform the shift+f shortcut on load), but should do the following:

if fitting to zoom will take the user to > 100%, limit the zoom to 100% and center the graph (i.e. zoom to fit, and then zoom to 100%)
image

if zooming to fit is <= 100%, simply zoom to fit.
image

There may also be something to say about setting a minimum zoom so you can't infinitely zoom out but this is an edge case.

ire4ever1190:
Closes #258 

Centres the graph when loading the project. Uses the rules laid out in the original issue so that we don't zoom out too far if the project is big


LvandoApps (commented):
Looks good, could have problems if the user creates multiple machines across the current project for some reason, but it's a very obscure case and likely will not happen, plus is only a UI feature.

LvandoApps (approved):
Approved +1 👍

PR #336, opened 2023-04-11
Start typing the event handling system

ire4ever1190:
This makes the event handling system be typed which has multiple benefits
- No need to specify parameter types in handler, they are now inferred (Cleans up the code nicely)
- Type safety in our handlers

This doesn't add types for every event since there is a lot of events that are still in JS files, but means that when we convert them to typescript we will start getting errors about the types being wrong (So we can do this incrementally without accidentally leaving them behind)


frontend/src/hooks/useEvent.ts
```diff
@@ -0,0 +1,38 @@
...
+export interface Events {
+  'editTransition': {id: number},
+  'editComment': {id: number, x: number, y: number},
+  'editStateName': {id: number},
+  'editStateLabel': {id: number},
+}
+
+interface EventOptions {
+  target: Document,
+  options?: any // boolean | AddEventListenerOptions
```
ire4ever1190, line 22: I don't really know what options is meant to be, I tried to find a place it is used in the code but couldn't yet.
I tried using the type specified by typescript for `addEventListener` but the types it uses aren't exported so I can't use them

ire4ever1190: Never mind found a few uses, will clean it up

SidhraFernando-Plant (approved):
Awesome job on this, it's fantastic that we can add them as we go without breaking everything 🔥

PR #338, opened 2023-04-13
Implement copy/paste feature

SidhraFernando-Plant:
### Overview
Implement an initial version of Copy/Paste. Not yet perfect, but it will be further refined as I work on template feature (#319).

### Solution
**Copy works like this:**
- Grab all of the states, transitions, and comments currently selected from useSelectionStore
- Put them into a JSON object called copyData
- Stringify copyData and copy to the clipboard with navigator.clipboard.writeText( )
  - writeText( ) and readText( ) were used as it seems as if clipboard.read( ) and clipboard.write( ) have more limited browser support

**Paste works like this:**
- Get the copied data as JSON by parsing what's in the clipboard - this is called pasteData
- Check whether user is copying to the same type of project
- Add all the states from pasteData and then select them, keeping track of their new ids along the way
  - The transitions are updated to link with the new state ids in this process
- Add all the comments and then select them
- Add all of the transitions and then select them
- Update the initialState if an initial state was copied

### Behaviour specifics
- An error appears if a user tries to paste something into a different project type
- An initial state will be copied as a normal state if the project it is copied into already has an initial state
- An error will occur if a user tries to copy a transition without copying both of the states it is linked to (ie. to and from) - error is not currently handled so nothing will happen or app will crash

### Known issues and future improvements
- It is not guaranteed that elements will paste in view, this should be implemented ASAP. It's mainly an issue if pasting between different projects
- Better handling of a situation where a transition is copied without its associated states should be implemented. It should work fine when pasting in the same project at the very least, and a better error message should be shown
- The paste function is really long, the code will likely be tidied up when templates are implemented as there will be shared code

### DIFF
--- a/frontend/src/hooks/useActions.js
+++ b/frontend/src/hooks/useActions.js
@@ -2,7 +2,7 @@ import { useEffect, useMemo } from 'react'
 import { useNavigate } from 'react-router-dom'
 
 import { useProjectsStore, useProjectStore, useSelectionStore, useToolStore, useViewStore } from '/src/stores'
-import { SCROLL_MAX, SCROLL_MIN, VIEW_MOVE_STEP } from '/src/config/interactions'
+import { SCROLL_MAX, SCROLL_MIN, VIEW_MOVE_STEP, COPY_DATA_KEY } from '/src/config/interactions'
 import { PASTE_POSITION_OFFSET } from '/src/config/rendering'
 import { convertJFLAPXML } from '@automatarium/jflap-translator'
 import { haveInputFocused } from '/src/util/actions'
@@ -150,13 +150,17 @@ const useActions = (registerHotkeys = false) => {
           projectType: project.projectType,
           initialStateId: isInitialSelected ? project.initialState : null
         }
-        navigator.clipboard.writeText(JSON.stringify(copyData))
+        sessionStorage.setItem(COPY_DATA_KEY, JSON.stringify(copyData))
       }
     },
     PASTE: {
       hotkey: { key: 'v', meta: true },
-      handler: async () => {
-        const pasteData = JSON.parse(await navigator.clipboard.readText())
+      handler: () => {
+        const pasteData = JSON.parse(sessionStorage.getItem(COPY_DATA_KEY))
+        if (pasteData === null) {
+          // Copy has not been executed
+          return
+        }
         let isInitialStateUpdated = false
         if (pasteData.projectType !== project.projectType) {
           alert(`Error: you cannot paste elements from a ${pasteData.projectType} project into a ${project.projectType} project.`)
@@ -215,6 +219,7 @@ const useActions = (registerHotkeys = false) => {
         if (isNewProject && pasteData.initialStateId !== null && project.initialState === null) {
           setStateInitial(pasteData.initialStateId)
         }
+        commit()
       }
 
     },
--- a/frontend/src/hooks/useActions.js
+++ b/frontend/src/hooks/useActions.js
@@ -191,6 +191,12 @@ const useActions = (registerHotkeys = false) => {
           }
           state.id = newId
         })
+        // Error if trying to paste transition without its to and from states
+        if (newTransitions.find(transition => transition.from === null || transition.to === null)) {
+          alert('Sorry, there was an error while pasting')
+          removeStates(pasteData.states.map(state => state.id))
+          return
+        }
         pasteData.transitions = newTransitions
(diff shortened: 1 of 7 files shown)


frontend/src/hooks/useActions.js
```diff
@@ -114,12 +123,100 @@ const useActions = (registerHotkeys = false) => {
...
+          alert(`Error: you cannot paste elements from a ${pasteData.projectType} project into a ${project.projectType} project.`)
+          return
+        }
+        const isNewProject = pasteData.projectSource !== project._id
+        // Track which transitions have been updated with new state ids
+        const newTransitions = structuredClone(pasteData.transitions)
+        newTransitions.forEach(transition => {
+          transition.from = null
+          transition.to = null
+        })
```
SidhraFernando-Plant, line 171: I ran into a rare bug when the transitions were being updated with the new state ids of the newly created states where the ids would get updated multiple times. So I had to have a way of marking which ids had been updated and which weren't to avoid updating ids more than once. The duplicate array of transitions starts as all null and then ids are updated, so we know that if an id is not null it has already been updated to the new state id and shouldn't be touched. We still have the original transitions in pasteData so that we know what needs to be updated. If this is not clear, I am happy to explain further with an example or discuss alternative approaches because I feel like there's probably better ways to do this

frontend/src/hooks/useActions.js
```diff
@@ -114,12 +123,100 @@ const useActions = (registerHotkeys = false) => {
...
+            }
+            if (transition.to === state.id && newTransitions[i].to === null) {
+              newTransitions[i].to = newId
+            }
+          })
+          // Update initial state id if applicable
+          if (pasteData.initialStateId === state.id && !isInitialStateUpdated) {
+            pasteData.initialStateId = newId
+            isInitialStateUpdated = true
+          }
```
SidhraFernando-Plant, line 191: Similarly to the transitions, we must only update the initialStateId to correspond the newly created state's ID once or we run into issues.

ire4ever1190 (changes requested):
Looking good :rocket: 

The issue about states not being guaranteed to be pasted in view, I noticed you have some offset code and wondering this means that the paste might be cut off a bit if its too large or if that the offset isn't done correctly? If its that large pastes are a bit cut off then imo that isn't an issue

But also bad news :disappointed:  this doesn't work in firefox since only [browser extensions are allowed to read the clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility). I did look into it and found another solution where you handle the [paste](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event) event and then check that the target is the svg element

SidhraFernando-Plant:
> The issue about states not being guaranteed to be pasted in view, I noticed you have some offset code and wondering if this means that the paste might be cut off a bit if its too large or if that the offset isn't done correctly? If its that large pastes are a bit cut off then imo that isn't an issue

I mentioned this in discord but also putting it here, the issue with the pasting in view mainly concerns pasting between projects. Say I copy a state from one project, and = the x value is 10 and the y value is 10. If I now go into a different project the view window in that project may be completely different from that of the last project, and x 10 and y10 may be completely out of view. So when I paste the state it will be out of view. The useViewStore will have to be used to ensure the x and y coordinates of the object to be pasted are in view.

frontend/src/hooks/useActions.js
```diff
@@ -114,12 +123,105 @@ const useActions = (registerHotkeys = false) => {
...
+        // This will use the CopyData type defined in ProjectTypes
+        const copyData = {
+          states: selectedStates,
+          comments: selectedComments,
+          transitions: selectedTransitions,
+          projectSource: project._id,
+          projectType: project.projectType,
+          initialStateId: isInitialSelected ? project.initialState : null
+        }
+        sessionStorage.setItem(COPY_DATA_KEY, JSON.stringify(copyData))
```
ire4ever1190, line 153: Think `localStorage` might be more appropriate?

One problem I found with `sessionStorage` when testing the branch was that [sessionStorage is per tab and doesn't really sync well if you have two tabs open](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API#web_storage_concepts_and_usage). Imo the times I'd want to be copying/pasting between projects I'd have multiple tabs open so would be good to have it work in those cases

SidhraFernando-Plant: My bad, I got ahead of myself and forgot to test again between tabs

ire4ever1190 (approved):

PR #339, opened 2023-04-13
Remove the delete tool

ire4ever1190:
We decided that it wasn't a good idea to have the delete tool cause of UX (You can already right click)


tdib (changes requested):
Should also remove the associated animation so there aren't any unnecessary files

GRA0007 (approved):
Keyboard shortcut dialog is broken?

tdib (changes requested):
The keyboard shortcut "d" deselects any tools - the shortcut should also be removed :)

ire4ever1190:
> Keyboard shortcut dialog is broken?

Separate issue, looking into it now

PR #341, opened 2023-04-14
Update `zustand`

### LINKED ISSUE #329: Update to latest zustand

The zustand version we use is a major version behind (`v3` vs `v4`). `v4` has better support for typescript and fixes the hack were we skip lib checking in the type checking.

Simply bumping the dependency is not enough since some things are deprecated and a few files were broken when I did

### Alternatives

_No response_

### Additional information

_No response_

ire4ever1190:
Closes #329 

This updates zustand to 4.x.x which brings better typescript support. This means we can reenable lib checking which stronger type checking

Turns out the issues that needed to be fixed were a lot less than what the red lines in my IDE were saying

### DIFF
--- a/frontend/src/stores/usePreferencesStore.ts
+++ b/frontend/src/stores/usePreferencesStore.ts
@@ -1,4 +1,4 @@
-import create, { SetState } from 'zustand'
+import { create, SetState } from 'zustand'
 import { persist } from 'zustand/middleware'
 
 interface Preferences {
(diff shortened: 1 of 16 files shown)


frontend/src/stores/usePreferencesStore.ts
```diff
@@ -20,7 +20,7 @@ const defaultPreferences: Preferences = {
   ctrlZoom: !navigator.platform?.match(/Win/) // Default to false on windows, which more often has a mouse
 }
 
-const usePreferencesStore = create<PreferencesStore>(persist((set: SetState<PreferencesStore>) => ({
+const usePreferencesStore = create<PreferencesStore>()(persist((set: SetState<PreferencesStore>) => ({
```
ire4ever1190, line 23: Some of the creates need to have an extra function call like [described here](https://github.com/pmndrs/zustand/issues/1013#issuecomment-1203781863)

LvandoApps (approved):
Perfect

Issue #242, opened 2022-07-12
Transition arrow overlap on vertical state positioning


tdib:
When states are positioned one over the other and there are multiple transitions between them, they overlap and make it impossible to read
![image](https://user-images.githubusercontent.com/79500311/178514093-424212da-a64c-4a10-b2b0-79ff84e88c43.png)

(for reference - here are the actual transitions)
![image](https://user-images.githubusercontent.com/79500311/178514280-5d7cde0b-b63d-45ee-8b5d-f02bc971a265.png)

PR #342, opened 2023-04-15
Make transition labels stack on top of each other

### LINKED ISSUE #281: Transition label stacking
Currently when two states have multiple transitions between them, each transition is a seperate arrow. It would be ideal if the transitions all stacked onto one arrow. Self transitions already do this. Particularly important due to the amount of transitions that can be required between two states in Computing Theory assignments.

Current behaviour:
Screen Shot 2023-03-12 at 9 56 03 pm

Desired behaviour:
Screen Shot 2023-03-12 at 9 57 00 pm

Current behaviour of self transitions (reflects desired behaviour):
Screen Shot 2023-03-12 at 9 59 54 pm

ire4ever1190:
closes #281, #242

image

Transition labels are now stacked on top of each other instead of creating a new edge. This makes the behaviour be like JFLAP where the transitions labels are like a column that are aligned to the normal of the edge.

Selecting a transition can be done by just clicking the label, if you want to select all transitions then you can select the edge.

This also converts a few more files to typescript that I was interacting with


SidhraFernando-Plant (changes requested):
sorry to be the spacing police again 🚨🚔🚨 but I feel there is a bit too much space now between the transitions and the labels (ie. transition arrow and first transition).

Screen Shot 2023-04-17 at 1 04 32 pm

 If possible, it would be good to reduce this spacing closer to what it was before. The spacing between the stacked transitions themselves is fine, though it would be fine to reduce this a bit if needed to bring the labels closer to the transition.

SidhraFernando-Plant:
Everything else looks great though, and it works really smoothly 👌

ire4ever1190:
All good, always feel free to be spacing police :muscle: 

Should be easy fix, will do :+1:

SidhraFernando-Plant (changes requested):
Looking good for transitions between different states ⭐️ but the spacing for self-transitions is still at the larger value - it would be good to bring this down to the same as the other transitions


Screen Shot 2023-04-19 at 7 28 54 pm

ire4ever1190:
image
The spacing has been changed and is a lot better now implementation wise. I now properly calculate the coordinate of the midpoint and then offset from that which is way less hacky than what I was doing before

SidhraFernando-Plant (approved):
Great job refining this solution 👍🏻

PR #343, opened 2023-04-15
Fix loading no project and convert `Editor` to typescript

ire4ever1190:
This converts `Editor` to typescript and fixes the bug where the editor would fail to load if a project wasn't already selected (e.g. going directly to `/editor` on first load).

The conversion to typescript required adding support for DOM events when using `useEvent` which is achieved by creating a union between our custom events and the DOM events


SidhraFernando-Plant (approved):
Typescript wizard 🧙🏻‍♂️✨nice work

PR #344, opened 2023-04-17
Fix height differences in new project cards

### LINKED ISSUE #299: Fix discrepancies with heights of NewProjectCard elements
If the window becomes too small, new file descriptions take up more lines. This leads to some NewProjectCard elements having different height to those with less lines of text description. They need to all be the same height, regardless of how many lines the text takes up.

image

ire4ever1190:
Closes #299 

image

Reason I'm doing the issue is because I was only able to reproduce on my laptop

The root cause seemed to be font sizes. My chrome had a smaller font size so it could fit everything but my Firefox uses my system font which is a bit bigger so took more space in the first card. This makes the cards be based relative to the size of the fonts and should allow for roughly ~5 lines of text per card.

Requested multiple reviewers so that it can be tested on different font sizes and computers to make sure the changes don't cause weirdness for anyone else

### DIFF
--- a/frontend/src/pages/NewFile/components/NewProjectCard/newProjectCardStyle.ts
+++ b/frontend/src/pages/NewFile/components/NewProjectCard/newProjectCardStyle.ts
@@ -1,12 +1,13 @@
 import { styled } from 'goober'
 
-export const CardContainer = styled('button')`
+export const CardContainer = styled('button')<{height: number}>`
   display: grid;
   grid-template-rows: 1.3fr 1fr;
   box-sizing: border-box;
   min-width: 18em;
   width: 20em;
-
+  height: ${p => p.height || '100%'}px;
+  
   border-radius: .5rem;
   border: 3px solid transparent;
   overflow: hidden;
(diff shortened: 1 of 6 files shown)


SidhraFernando-Plant (dismissed):
I am a little concerned about the added space after the text:

Screen Shot 2023-04-17 at 12 18 51 pm

A few thoughts:
- Is there any way the values can be adjusted to prevent this?
- More space at the bottom would probably be fine as long as the space at the top is the same, ie. text is vertically centred

LvandoApps (changes requested):
Yeah formatting is off

ire4ever1190:
In the end I just calculated the height for the tallest card and set the rest to that.

Seemed to be the easiest solution since it means no fiddling with fonts, css, etc
Also means it is future proof and wont need changing if any text is added/removed

frontend/src/pages/NewFile/components/NewProjectCard/newProjectCardStyle.ts
```diff
@@ -1,12 +1,13 @@
...
 
-export const CardContainer = styled('button')`
+export const CardContainer = styled('button')<{height: number}>`
   display: grid;
   grid-template-rows: 1.3fr 1fr;
   box-sizing: border-box;
   min-width: 18em;
   width: 20em;
-
+  height: ${p => p.height || '100%'}px;
```
ire4ever1190, line 9: In case there is a problem in the state, we just set the height to 100% so its least usable

LvandoApps (approved):
Good!

PR #347, opened 2023-04-20
NFA to DFA converter

LvandoApps:
This is the finished version of the NFA to DFA converter, currently it does not have an algorithm to reorganise the graph such that the states are laid out properly (this is an issue that will be done separately in the future) so for now just drag each state to where you would like it to be, however other than this all other required features are incorporated.

You can test it by creating an FSA project and going "Tools > Convert NFA to DFA", which will create all the states in one position as mentioned earlier. If you have an incorrect graph nothing will happen (such as not having an initial state, final state, those states aren't reachable, etc), this is intentional. 

Here is an example:

NFA:
Screenshot 2023-04-20 172210

DFA:
dfaStuff

As can be seen it has converted correctly and is following all the rules that a DFA must follow.


frontend/src/hooks/useActions.js
```diff
@@ -199,7 +201,13 @@ const useActions = (registerHotkeys = false) => {
...
     },
     CONVERT_TO_DFA: {
-      // handler: () => console.log('Convert to DFA'),
+      handler: projectType === 'FSA'
+        ? () => {
+            updateProject(reorderStates(convertNFAtoDFA(reorderStates(project))))
+            console.log(reorderStates(convertNFAtoDFA(reorderStates(project))))
+            commit()
+          }
+        : undefined
```
ire4ever1190, line 210:
```suggestion
      handler: () => {
            if (projectType !== 'FSA') return
            updateProject(reorderStates(convertNFAtoDFA(reorderStates(project))))
            console.log(reorderStates(convertNFAtoDFA(reorderStates(project))))
            commit()
          }
```
Currently working on converting `useActions` to typescript and undefined functions wont be allowed.
There is also a disabled field that you can use to toggle making the action disabled

packages/simulation/tests/convertNFAtoDFA.test.ts
ire4ever1190, line 1: Would be good to group all the tests under a suite (using `describe`)

ire4ever1190: Would be good to also strength some of the tests and check that the transitions have the expected symbols

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+  return dfaGraph
+}
+
+// This will check to ensure that the graph passed in has valid states/transitions before continuing
+export const statesAndTransitionsPresent = (nfaGraph: FSAGraphIn): boolean => {
+  if (nfaGraph.states.length === 0 || nfaGraph.transitions.length === 0) {
+    return false
+  } else {
+    return true
+  }
```
ire4ever1190, line 77:
```suggestion
  return !(nfaGraph.states.length === 0 || nfaGraph.transitions.length === 0)
```

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+}
+
+// This will check to ensure that the graph passed in has an initial state before continuing
+export const initialStateIsPresent = (nfaGraphInitialState: StateID): boolean => {
+  if (nfaGraphInitialState == null) {
+    return false
+  } else {
+    return true
+  }
```
ire4ever1190, line 102:
```suggestion
  return nfaGraphInitialState != null
```

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+}
+
+// This will check if the merged state is different from the current state's merged state
+function arraysEqual (arr1, arr2) {
```
ire4ever1190, line 106: We already have the lodash library so best to use the [isEqual](https://lodash.com/docs/4.17.15#isEqual) function from that

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+  for (let curElem = 0; curElem < numberOfNFAStates; curElem++) {
+    if (nfaGraphStates[curElem].isFinal) {
+      doesFinalExist = true
+    }
+  }
+  if (!doesFinalExist) {
+    return false
+  } else {
+    return true
+  }
```
ire4ever1190, line 93:
```suggestion
  return nfaGraphStates.some(state => state.isFinal)
```

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+    }
+    if (!symbolFound) {
+      return false
+    }
+  }
+  return true
+}
+
+// This function will handle the logic for combining states when necessary
+function combineStates (initialTransitionTable: {[key: StateID]: [StateID, ReadSymbol][]}, symbolsArray: ReadSymbol[], curTransitionElem: number, symbolToStatesMap: {[key: ReadSymbol]: [StateID, StateID][]}, symbol: ReadSymbol): [string, { symbol: ReadSymbol, tostates: StateID[] }[]] {
```
ire4ever1190, line 136: Make a type alias for the `initialTransitionTable` at the start of the file and use that
```ts
type TransitionTable = Record<StateID, [StateID, ReadSymbol][]>
```
I also notice a lot of loops to find states, might it be better to have it be `Record<StateID, Record<StateID, ReadSymbol>>` or is the algorithm easier this way?

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+    }).sort()
+
+    return { symbol, tostates }
+  })
+
+  return [curKeyCombination, combinedState]
+}
+
+// This will mimic STEP 1 of the procedure by ensuring that for every DFA state, there are no lambda transitions and the states are merged correctly
+export function removeLambdaTransitions (initialTransitionTable: {[key: StateID]: [StateID, ReadSymbol][]}, mergedStates: {[key: string]: StateID[]}, numberOfNFAStates: StateID, curInitialState: StateID, curFinalStates: StateID[], tempTransitionTable: {[key: StateID]: [StateID, ReadSymbol][]}): [{[key: StateID]: [StateID, ReadSymbol][]}, StateID, StateID[]] {
```
ire4ever1190, line 161: Same here, these types seem to be repeated in a lot of places so best to move them into actual types

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+          }
+        } else if (!hasTransitions && nfaGraph.states[stateID].isFinal) {
+        // Else if the state is a final state but still needs transitions, it can just put all those transitions into one trap state
+          initialTransitionTable[nextAvailableStateID] = []
+          // New trap state
+          for (let curElem = 0; curElem < symbolsArray.length; curElem++) {
+            initialTransitionTable[nextAvailableStateID].push([nextAvailableStateID, symbolsArray[curElem]])
+          }
+          // Transitions to the trap state
+          for (let curElem = 0; curElem < symbolsArray.length; curElem++) {
```
ire4ever1190, line 328: There are a lot of loops like this, but the index is only used to access the array, why not just use a `for of` loop?

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+  const curFinalStates: StateID[] = result[2]
+  dfaGraph.initialState = curInitialState
+  for (let curObject = 0; curObject < Object.keys(transitionTable).length; curObject++) {
+    dfaGraph.states[curObject] = { id: undefined, isFinal: undefined, x: 765, y: 330 }
+    dfaGraph.states[curObject].id = parseInt(Object.keys(transitionTable)[curObject])
+    if (curFinalStates.includes(dfaGraph.states[curObject].id)) {
+      dfaGraph.states[curObject].isFinal = true
+    } else {
+      dfaGraph.states[curObject].isFinal = false
+    }
```
ire4ever1190, line 596:
```suggestion
    dfaGraph.states[curObject].isFinal = curFinalStates.includes(dfaGraph.states[curObject].id)
```

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+  const lastKey = Object.keys(initialTransitionTable)[Object.keys(initialTransitionTable).length - 1]
+  let lastKeyNumber = parseInt(lastKey)
+  const lastKeyArray = []
+  // Increase by one to get to next key (the new key for initial table)
+  lastKeyNumber += 1
+  for (let mergedElem = 0; mergedElem < Object.keys(newCombinedStates).length; mergedElem++) {
+    const stateArray = Object.keys(newCombinedStates)[mergedElem].split(',').map(Number)
+    for (let curState = 0; curState < numberOfNFAStates; curState++) {
+      if (initialTransitionTable[curState] === undefined) {
+        continue
```
ire4ever1190, line 467: Just have it as `if (initialTransitionTable[curState] !== undefined)` instead of this `if/else`

ire4ever1190: I notice this idiom is used in a few places, make sure to clean it up in the other places

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+
+// Define a type for the DFAGraph
+type DFAGraph = {
+  initialState: StateID;
+  states: AutomataState[];
+  transitions: AutomataTransition[];
+}
+
+// Graph traversal algorithm starting from the initial state that removes unreachable states and transitions
+export const removeUnreachableNFAStates = (nfaGraph: FSAGraphIn): FSAGraphIn => {
```
ire4ever1190, line 22: Take a look at closureWithPredicate.ts, it basically implements the graph exploration you have here so you can use that instead of reinventing the wheel. Should work for `removeUnreachableDFAStates` also (or atleast should once I finish cleaning the types on it)

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
+import { FSAGraphIn, FSAState, StateID, ReadSymbol } from './graph'
+import { AutomataState, AutomataTransition } from 'frontend/src/types/ProjectTypes'
+
+// EXTREMELY IMPORTANT NOTE:
```
ire4ever1190, line 4: Are these notes still valid? Thought you fixed the sequential issue?

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+      // Sort the states so that they're ordered. This is important as the keys of the objects is used in the conversion algorithm
+      nfaGraph.states.sort((a, b) => a.id - b.id)
+      // Create a DFA from the given NFA
+      dfaGraph = createDFA(nfaGraph, dfaGraph)
+      // Remove unreachable states from this new dfaGraph
+      dfaGraph = removeUnreachableDFAStates(dfaGraph)
+      // Finally copy over the required elements from nfaGraph and return the final dfaGraph
+      const { initialState, states, transitions, ...dfaGraphCopy } = nfaGraph
+      const dfaGraphFinal = { ...dfaGraph, ...dfaGraphCopy }
+      return dfaGraphFinal
```
ire4ever1190, line 644:
```suggestion
      return { ...dfaGraph, ...dfaGraphCopy }
```
No need to make a temp

ire4ever1190 (changes requested):
Looking good :rocket: 
There is a lot of weird stuff in `convert.ts` though, I mentioned the first time the items popped up but a lot of the issues are repeated so much sure to check other places and not just where I left the note.

Also where are you following the algorithm from? Would be good to know what the source of it is

ire4ever1190:
Just a quick review, will try and look further at the conversion code so I can give a more indepth review :+1:

Good work on implementing it though, we are slowing catching up to JFLAP in these complex features :muscle:

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,647 @@
...
+      return false
+    }
+  }
+  return true
+}
+
+// This will check that a given states transitions contain all the symbols
+function hasAllSymbols (transitions, symbols) {
+  for (const symbol of symbols) {
+    let symbolFound = false
```
ire4ever1190, line 121: Can use `transitions.some(...)` here also

Also make sure to type the function

packages/simulation/src/convert.ts
LvandoApps: I'll just leave it for now then if it isn't compatible with DFA graph as that's an important part of the algorithm

LvandoApps:
> Looking good 🚀 There is a lot of weird stuff in `convert.ts` though, I mentioned the first time the items popped up but a lot of the issues are repeated so much sure to check other places and not just where I left the note.
> 
> Also where are you following the algorithm from? Would be good to know what the source of it is

My brain, that's why it took so long haha, all of it is from scratch just applying prior knowledge I know on the NFA to DFA conversion logic. The best way of doing that is through a transition table, which is the basis of my algorithm (working with this transition table).

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,571 @@
...
+// This will check to ensure that the graph passed in has a final state before continuing
+export const finalStateIsPresent = (nfaGraphStates: FSAState[]): boolean => {
+  const numberOfNFAStates: number = nfaGraphStates.length
+  let doesFinalExist: boolean = false
+  for (let curElem = 0; curElem < numberOfNFAStates; curElem++) {
+    if (nfaGraphStates[curElem].isFinal) {
+      doesFinalExist = true
+    }
+  }
+  return doesFinalExist
```
ire4ever1190, line 75:
```suggestion
  return nfaGraphStates.some(state => state.isFinal)
```

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,571 @@
...
+    if (foundInitial && foundFinals) {
+      break
+    }
+  }
+
+  // Merge the states in tempTransitionTable and initialTransitionTable based on the mergedStates array. Go top down so that states that are valid states are also
+  // able to transition to the last state in the chain
+  for (let curElem = numberOfNFAStates - 1; curElem >= 0; curElem--) {
+    const mergedWith = mergedStates[curElem][0]
+    for (let i = 1; i < mergedStates[curElem].length; i++) {
```
ire4ever1190, line 176: Please look through your for loops again

`i` is only used on the next line to index `mergedStates[curElem]` so this could just be a `for of` loop

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,571 @@
...
+              } else {
+                existingTransition[0] = nextAvailableStateID
+              }
+            }
+          }
+          nextAvailableStateID++
+        } else {
+        // Else go each transition and if a symbol is not found, make a new state for it and transition to that state with the given symbol.
+          for (const symbol of symbolsArray) {
+            let symbolFound = false
```
ire4ever1190, line 282: `some` could also be used here

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,571 @@
...
+      })
+    }
+    symbolToStatesMap = {}
+    // The from state is just temporary and is not important, so use to identify the sequences
+    tempFromState = 0
+    for (const [key, value] of Object.entries(newCombinedStates)) {
+      for (let transElem = 0; transElem < value.length; transElem++) {
+        if (newCombinedStates[key][transElem].symbol in symbolToStatesMap) {
+          for (let toStateElem = 0; toStateElem < newCombinedStates[key][transElem].tostates.length; toStateElem++) {
+            symbolToStatesMap[newCombinedStates[key][transElem].symbol].push([tempFromState, newCombinedStates[key][transElem].tostates[toStateElem]])
```
ire4ever1190, line 370: `newCombinedStates[key][transElem]` is accessed a lot, would be good to move it into a variable

packages/simulation/src/convert.ts
```diff
@@ -0,0 +1,571 @@
...
+          const filtered = initialTransitionTable[curState].filter(obj => obj[1] === symbolsArray[curSymbolElem])
+          // In this case there is a chance that these transitions are what need to be deleted as they have been merged
+          if (filtered.length > 1 && filtered.length === stateArray.length) {
+            filtered.sort((a, b) => a[0] - b[0])
+            let sameTransitions = true
+            for (let checkElem = 0; checkElem < filtered.length; checkElem++) {
+              if (filtered[checkElem][0] !== stateArray[checkElem]) {
+                sameTransitions = false
+              }
+            }
```
ire4ever1190, line 408:
```suggestion
            const sameTransitions = !filtered.some((item, i) => item[0] !== stateArray[i])
```
`some` could also be used here like this. My logic might be wrong so double check

frontend/src/hooks/useActions.js
```diff
@@ -199,7 +201,12 @@ const useActions = (registerHotkeys = false) => {
       handler: () => dispatchCustomEvent('sidepanel:open', { panel: 'options' })
     },
     CONVERT_TO_DFA: {
-      // handler: () => console.log('Convert to DFA'),
+      handler: () => {
+        if (projectType !== 'FSA') return
+        updateProject(reorderStates(convertNFAtoDFA(reorderStates(project))))
+        console.log(reorderStates(convertNFAtoDFA(reorderStates(project))))
```
ire4ever1190, line 207: Make sure to remove the debug statement

ire4ever1190 (changes requested):
Now this is not an exhaustive list of everything since there are quite a few places that the loops are still weird so please do look through your whole code

I still don't fully understand the algorithm and still trying to work my way through it

packages/simulation/src/convert.ts
ire4ever1190: Would the alternate be true? 
Like remove the double negation and make it  `filtered.some((item, i) => item[0] === stateArray[i])` cause that reads better imo

LvandoApps: Nah they're not the same, the original code and your first suggestion both ensure that the filtered array and the state array have the same to states (assuming they're sorted), if even one is different then it has to return false because it isn't a state that needs to be merged. The second only checks to see if one pair is the same rather than all of them.

ire4ever1190: Ah I see, so you'd need to use `every` instead of `some`

Alright thanks 👌

frontend/src/hooks/useActions.js
```diff
@@ -199,7 +201,11 @@ const useActions = (registerHotkeys = false) => {
       handler: () => dispatchCustomEvent('sidepanel:open', { panel: 'options' })
     },
     CONVERT_TO_DFA: {
-      // handler: () => console.log('Convert to DFA'),
+      handler: () => {
```
ire4ever1190, line 204:
```suggestion
      disabled: () => projectType !== 'FSA',
      handler: () => {
```

ire4ever1190: Used it in #349 and remembered I didn't see it here

LvandoApps: Ah so instead of checking if the project type is FSA I can just add that disabled "flag" instead?

ire4ever1190: ye, and then it will disable itself in the UI

packages/simulation/tests/convertNFAtoDFA.test.ts
```diff
@@ -0,0 +1,171 @@
...
+import convertSimpleConversion from './graphs/convertSimpleConversion.json'
+import convertHarderConversion from './graphs/convertHarderConversion.json'
+import convertMultipleFinal from './graphs/convertMultipleFinal.json'
+import convertInitialNotAtStart from './graphs/convertInitialNotAtStart.json'
+
+describe('Check to ensure NFA graph is valid before conversion begins', () => {
+  test('Graph should not be processed for conversion if there are no final states', () => {
+    try {
+      reorderStates(convertNFAtoDFA(reorderStates(convertFinalNotPresent as any) as any) as any)
+      fail('Expected an error to be thrown')
```
ire4ever1190, line 16: Use [.toThrow()](https://jestjs.io/docs/expect#tothrowerror) instead

LvandoApps: Will add now

ire4ever1190 (approved):
:+1:

Would be good to do a refactor sometime in the future now that you have the whole algorithm pieced together

PR #348, opened 2023-04-21
!DRAFT! Templates

SidhraFernando-Plant:
# DRAFT
Still working on this


frontend/src/hooks/useCreateBatch.ts
```diff
@@ -0,0 +1,78 @@
...
+    alert('Sorry, there was an error')
+    removeStates(createData.states.map(state => state.id))
+    return
+  }
+  createData.transitions = newTransitions
+  selectStates(createData.states.map(state => state.id))
+  createData.comments.forEach(comment => {
+    // TODO: ensure position isn't out of window
+    comment.x += PASTE_POSITION_OFFSET
+    comment.y += PASTE_POSITION_OFFSET
```
LvandoApps, line 62: Small nitpick, can replace this with the below:
`comment.x, comment.y = comment.x + PASTE_POSITION_OFFSET, comment.y + PASTE_POSITION_OFFSET`
See it a few times, nothing major just a fussy edit

SidhraFernando-Plant: Good point, will change this

frontend/src/hooks/useActions.js
```diff
@@ -563,4 +504,39 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
   input.click()
 }
 
+// Takes in the IDs of states, comments, and transitions
+// Parameters also include  the current project and whether a template is being created
+// Outputs a CopyData to be copied or Template object to be created into a template
+const selectionToCopyTemplate = (stateIds, commentIds, transitionIds, project, isTemplate) => {
```
SidhraFernando-Plant, line 510: This method takes a selection and makes it into a CopyData or Template for the COPY or CREATE_TEMPLATE actions

frontend/src/hooks/useCreateBatch.ts
```diff
@@ -0,0 +1,82 @@
...
+  const createComment = useProjectStore(s => s.createComment)
+  const createTransition = useProjectStore(s => s.createTransition)
+  const removeStates = useProjectStore(s => s.removeStates)
+  const selectStates = useSelectionStore(s => s.setStates)
+  const selectTransitions = useSelectionStore(s => s.setTransitions)
+  const selectComments = useSelectionStore(s => s.setComments)
+  const setStateInitial = useProjectStore(s => s.setStateInitial)
+  const commit = useProjectStore(s => s.commit)
+
+  const createBatch = () => {
```
SidhraFernando-Plant, line 18: I tried to make a hook for inserting + selecting a group of elements (states, transitions, comments) as this is used in the copy and template insert functions. I couldn't get it to work properly because I always had issues calling so I made it into a store method insertGroup in the useProjectStore. I would need some help/pairing to make this happen, but for now it works in the store although it doesn't seem like the best solution

SidhraFernando-Plant: (to clarify, this isn't actually called anywhere but I have left it in for now in case)

frontend/src/components/Sidepanel/Panels/Templates/Templates.js
ire4ever1190, line 1: Shouldn't this be in typescript?

frontend/src/hooks/useActions.js
```diff
@@ -41,17 +40,30 @@ const useActions = (registerHotkeys = false) => {
...
+  const insertGroup = useProjectStore(s => s.insertGroup)
+  const addTemplate = useTemplatesStore(s => s.upsertTemplate)
   const navigate = useNavigate()
 
   // TODO: memoize
   const actions = {
+    CREATE_TEMPLATE: {
+      hotkey: { key: 'j', meta: true },
+      handler: () => {
+        if (selectedStatesIds.length === 0 && selectedCommentsIds.length === 0 && selectedTransitionsIds.length === 0) {
```
ire4ever1190, line 56: Might be better to move this into the disabled field

frontend/src/hooks/useActions.js
```diff
@@ -563,4 +504,39 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
...
 
+// Takes in the IDs of states, comments, and transitions
+// Parameters also include  the current project and whether a template is being created
+// Outputs a CopyData to be copied or Template object to be created into a template
+const selectionToCopyTemplate = (stateIds, commentIds, transitionIds, project, isTemplate) => {
+  const selectedStates = stateIds.map((stateId) => {
+    return project.states.find((state) => {
+      return state.id === stateId
+    })
+  })
```
ire4ever1190, line 515: Imo it'll be cleaner to use `filter` instead of `map` and `find`
```suggestion
  const selectedStates = project.states.filter(state => stateIds.includes(state))
```

Could also move it into a simple helper like (This is in typescript so its clearer, I understand the file is still JS)
```suggestion
  const filterFrom = <T>(allowed: T[], from: T[]): T[] => from.filter(x => allowed.includes(x))
  const selectedComments = filterFrom(commentIds, project.comments)
```
Either way is fine, but make sure to update the comments and transitions filtering also

frontend/src/hooks/useTemplateInsert.ts
```diff
@@ -0,0 +1,66 @@
...
+  })
+  const offsetX = mousePos.x - originState.x
+  const offsetY = mousePos.y - originState.y
+  states.forEach((state) => {
+    state.x += offsetX
+    state.y += offsetY
+  })
+}
+
+const snapPosition = ({ x, y }) =>
```
ire4ever1190, line 63: This exact function is in `useStateCreation` also. Might be best to move the function into a util file like `points.js` instead of repeating it in multiple places

frontend/src/stores/useProjectStore.ts
```diff
@@ -24,11 +26,28 @@ import {
...
 
+import { PASTE_POSITION_OFFSET } from 'frontend/src/config/rendering.js'
+
 /**
  * A stored project has an extra `_id` field which is used to tell identify it
  */
 export type StoredProject = Project & {_id: string}
 
+export enum InsertGroupResponseType {
+  // eslint-disable-next-line no-unused-vars
```
ire4ever1190, line 37: Weird you need to disable, but I think I know the problem and will fix myself in a seperate PR :+1:

frontend/src/stores/useProjectStore.ts
```diff
@@ -226,6 +244,87 @@ const useProjectStore = create<ProjectStore>()(persist((set: SetState<ProjectSto
...
+        state.id = newId
+        // createState
+        project.states.push({ ...state })
+      })
+      createData.transitions = newTransitions
+      createData.comments.forEach((comment, i) => {
+        // TODO: ensure position isn't out of window
+        comment.x += PASTE_POSITION_OFFSET
+        comment.y += PASTE_POSITION_OFFSET
+        const newId = i + 1 + Math.max(-1, ...get().project.comments.map((c: ProjectComment) => c.id))
```
ire4ever1190, line 310: Move this out into a function like
```ts
const nextIDFor = (ids: {id?: number}[]): number => i + 1 + Math.max(-1, ...ids.map(s => s.id))
```
And then use in the other places that creating a new ID is needed. 

Also not too sure but looks like that during these loops where a newID is needed the only difference is `i +`? So might be better to cache the highest found and then just increment/add `i` inside the loop

ire4ever1190 (changes requested):
Looking good besides those few things :rocket: 

Like the UI concept :ok_hand: will be cool once that is finished in the next PR

frontend/src/components/Sidepanel/Panels/Templates/Templates.js
SidhraFernando-Plant: Oh yep sorry I was meant to update it, and then forgot 😅

frontend/src/stores/useProjectStore.ts
SidhraFernando-Plant: Oh yes I meant to single this out in a comment, linter wasn't liking enums 😢 Thank you, that would be great!

frontend/src/components/Toolbar/Toolbar.js
```diff
@@ -54,6 +54,14 @@ const tools = [
...
     icon: <MessageSquare />,
     animation: commentAnimation
+  },
+  {
+    label: 'Template tool',
+    hotkey: 'J',
+    description: 'Add templates to your automaton',
+    value: 'template',
+    icon: <Star />,
+    animation: commentAnimation
```
LvandoApps, line 64: Would be good if it had its own unique animation when you hover over the tool as I found this really useful in understanding how to use the other tools. I'm not familiar with how it's able to do that though I've never seen JSON files used in that way (see the JSON files in the animation folder in the Toolbar directory), might be best to consult group 1 on how they did this if neither you nor Jake know (assuming they implemented this animation feature).

LvandoApps:
Looking good so far, excited to see how the UI turns out 🚀

frontend/src/hooks/useActions.js
SidhraFernando-Plant: Is there a way I can use this command for testing if it's disable? If not, I'm going to leave it as not disabled for now so that I can keep testing it. By the time this is merged to dev the UI will hopefully be merged so it shouldn't need to be disabled (although if it wasn't finished when merging I would make sure it's disabled)

ire4ever1190: The logic will mostly stay the same, just means the UI button will be greyed out if nothing is selected (So you should be able to keep testing it)
```ts
disabled: () => selectedStatesIds.length === 0 && selectedCommentsIds.length === 0 && selectedTransitionsIds.length === 0,
handler: () => {
  // Normal code here, no need to check if stuff is selected anymore
```

SidhraFernando-Plant: Oh right, I thought you meant disable it entirely but that makes more sense

SidhraFernando-Plant:
I believe I have addressed all feedback, except disabling the action but I am going to do that along with the UI. 

I am going to branch off of this branch for the UI (I have merged with dev so this branch is not stale), then once everything is finalised I'll merge the whole branch train. So pretty much I am re-requesting review to check that I have suitably addressed all feedback, NOT with the intention of merging this branch to dev.

frontend/src/components/Sidepanel/Panels/Templates/templatesStyle.js
ire4ever1190, line 1: `.js -> .ts`

ire4ever1190 (approved):
Looks good, assuming you'll convert that `templatesStyle.js` during the UI part.

So is this PR staying open? Or close it and then open a new one once UI is finished?

frontend/src/components/Sidepanel/Panels/Templates/templatesStyle.js
SidhraFernando-Plant: Yep will convert to TS. I'll put this PR back to a draft I reckon til the other one is finished

ire4ever1190:
Re-requested myself so I don't forget 😵‍💫

Will review after you merge in dev :rocket:

frontend/src/hooks/useActions.ts
```diff
@@ -180,65 +159,17 @@ const useActions = (registerHotkeys = false) => {
...
-        })
-        selectComments(pasteData.comments.map(comment => comment.id))
-        pasteData.transitions.forEach(transition => {
-          const newId = createTransition(transition)
-          transition.id = newId
-        })
-        selectTransitions(pasteData.transitions.map(transition => transition.id))
-        if (isNewProject && pasteData.initialStateId !== null && project.initialState === null) {
-          setStateInitial(pasteData.initialStateId)
+        // createBatch(pasteData, project)
```
ire4ever1190, line 162:
```suggestion
```
Hook has been deleted iirc

frontend/src/hooks/useActions.ts
```diff
@@ -180,65 +159,17 @@ const useActions = (registerHotkeys = false) => {
...
+        const insertResponse = insertGroup(pasteData)
+        // This will be better in TS with enum
+        if (insertResponse.type === 2) {
```
ire4ever1190, line 165:
```suggestion
        if (insertResponse.type === InsertGroupResponseType.SUCCESS) {
```
Just so we aren't using a magical number

frontend/src/hooks/useActions.ts
```diff
@@ -180,65 +159,17 @@ const useActions = (registerHotkeys = false) => {
...
+          selectComments(insertResponse.body.comments.map(comment => comment.id))
+          selectStates(insertResponse.body.states.map(state => state.id))
+          selectTransitions(insertResponse.body.transitions.map(transition => transition.id))
+          commit()
+        } else if (insertResponse.type === 1) {
```
ire4ever1190, line 170:
```suggestion
        } else if (insertResponse.type === InsertGroupResponseType.FAIL) {
```
Same

frontend/src/hooks/useActions.ts
```diff
@@ -588,4 +522,23 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
   input.click()
 }
 
+// Takes in the IDs of states, comments, and transitions
+// Parameters also include  the current project and whether a template is being created
+// Outputs a CopyData to be copied or Template object to be created into a template
+export const selectionToCopyTemplate = (stateIds: number[], commentIds: number[], transitionIds: number[], project): CopyData => {
```
ire4ever1190, line 528:
```suggestion
export const selectionToCopyTemplate = (stateIds: number[], commentIds: number[], transitionIds: number[], project: StoredProject): CopyData => {
```
If you set the type here, then you won't need to specify the types later on when filtering

frontend/src/hooks/useActions.ts
```diff
@@ -588,4 +522,23 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
...
+  const isInitialSelected = stateIds.includes(project.initialState)
+  const returnObject: CopyData = {
+    states: selectedStates,
+    comments: selectedComments,
+    transitions: selectedTransitions,
+    projectSource: project._id,
+    projectType: project.projectType,
+    initialStateId: isInitialSelected ? project.initialState : null
+  }
+  return returnObject
```
ire4ever1190, line 541:
```suggestion
  return {
    states: selectedStates,
    comments: selectedComments,
    transitions: selectedTransitions,
    projectSource: project._id,
    projectType: project.projectType,
    initialStateId: isInitialSelected ? project.initialState : null
  }
```
Very small nitpick 😄 I just usually like to return objects directly

frontend/src/hooks/useCreateBatch.ts
ire4ever1190: If you don't need it anymore then feel free to delete 👍
We can recover it through git anyways if we need it later

frontend/src/stores/useProjectStore.ts
```diff
@@ -224,6 +245,94 @@ const useProjectStore = create<ProjectStore>()(persist((set: SetState<ProjectSto
...
+        }
+      })
+      // If a transition has to or from missing, flag to return error
+      if (!(toFound && fromFound)) {
+        missingState = true
+      }
+    })
+    if (missingState) {
+      return { type: InsertGroupResponseType.FAIL, body: 'Sorry, there was an error.' }
+    }
```
ire4ever1190, line 278:
```suggestion
    const stateIDs = new Set(createData.states.map(it => it.id))
    for (const transition of createData.transitions) {
      const hasTo = stateIDs.has(transition.to)
      const hasFrom = stateIDs.has(transition.from)
      // If either `to` or `from` is missing then return an error
      if (!(hasTo && hasFrom)) {
        return { type: InsertGroupResponseType.FAIL, body: 'Sorry, there was an error.' }
      }
    }
```
Using sets is a bit cleaner imo, lemme know what you think though.
I tired to do the same logic so should work the same (if something is fishy plz mention). Main difference is that we return the error inside the loop so that it breaks early if we know something is wrong

frontend/src/stores/useProjectStore.ts
```diff
@@ -224,6 +245,94 @@ const useProjectStore = create<ProjectStore>()(persist((set: SetState<ProjectSto
...
+        [comment.x, comment.y] = [comment.x + positionOffset, comment.y + positionOffset]
+        const newId = nextCommentId + i
+        comment.id = newId
+        // createComment
+        project.comments.push({ ...comment })
+      })
+      const nextTransitionId = nextIDFor(project.transitions)
+      createData.transitions.forEach((transition, i) => {
+        const newId = nextTransitionId + i
+        transition.id = newId
```
ire4ever1190, line 325:
```suggestion
        transition.id = nextTransitionId + i
```
Same

frontend/src/stores/useProjectStore.ts
```diff
@@ -224,6 +245,94 @@ const useProjectStore = create<ProjectStore>()(persist((set: SetState<ProjectSto
...
+        // createState
+        project.states.push({ ...state })
+      })
+      createData.transitions = newTransitions
+      const nextCommentId = nextIDFor(project.comments)
+      createData.comments.forEach((comment, i) => {
+        // TODO: ensure position isn't out of window
+        [comment.x, comment.y] = [comment.x + positionOffset, comment.y + positionOffset]
+        const newId = nextCommentId + i
+        comment.id = newId
```
ire4ever1190, line 318:
```suggestion
        comment.id = nextCommentId + i
```
Same small nitpick. Good in places like this though since it means one less variable needs to be kept in mental context when reading code

frontend/src/stores/useTemplateStore.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import {
+  Template
+} from '../types/ProjectTypes'
+
+interface TemplateStore {
+  template: Template,
+  setTemplate: (template: Template) => void
+  // update: (template: Template) => void,
+  // setName: (newName: string) => void
```
ire4ever1190, line 11:
```suggestion
  // TODO: Support updating/renaming
```
Just so we can find what needs to be done by searching for `// TODO`

frontend/src/stores/useTemplateStore.ts
```diff
@@ -0,0 +1,27 @@
...
+  template: null,
+  setTemplate: template => {
+    set({ template })
+  }
+  // update: (template) => {
+
+  // },
+  // setName: (newName) => {
+
+  // }
```
ire4ever1190, line 24:
```suggestion
```

frontend/src/hooks/useActions.ts
```diff
@@ -249,14 +180,17 @@ const useActions = (registerHotkeys = false) => {
...
-        const selectionState = useSelectionStore.getState()
-        const selectedStateIDs = selectionState.selectedStates
-        const selectedTransitionIDs = selectionState.selectedTransitions
-        const selectedCommentIDs = selectionState.selectedComments
-        if (selectedStateIDs.length > 0 || selectedTransitionIDs.length > 0 || selectedCommentIDs.length > 0) {
-          removeStates(selectedStateIDs)
-          removeTransitions(selectedTransitionIDs)
-          removeComments(selectedCommentIDs)
+        // If a template is selected and nothing else is, delete the template
+        if (template !== null && selectedCommentsIds.length === 0 && selectedStatesIds.length === 0 && selectedTransitionsIds.length === 0) {
```
ire4ever1190, line 184: Might be good to move out the length checks into a variable
```ts
const nothingSelected = selectedCommentsIds.length === 0 && selectedStatesIds.length === 0 && selectedTransitionsIds.length === 0
```
And then the else if can just be `else if (!nothingSelected)`

frontend/src/hooks/useActions.ts
```diff
@@ -588,4 +522,23 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
   input.click()
 }
 
+// Takes in the IDs of states, comments, and transitions
+// Parameters also include  the current project and whether a template is being created
+// Outputs a CopyData to be copied or Template object to be created into a template
+export const selectionToCopyTemplate = (stateIds: number[], commentIds: number[], transitionIds: number[], project): CopyData => {
+  const selectedStates: AutomataState[] = project.states.filter(state => stateIds.includes(state.id))
+  const selectedComments: ProjectComment[] = project.comments.filter(comment => commentIds.includes(comment.id))
```
ire4ever1190, line 530:
```suggestion
  const selectedStates = project.states.filter(state => stateIds.includes(state.id))
  const selectedComments = project.comments.filter(comment => commentIds.includes(comment.id))
```
No need after previous suggestion

frontend/src/stores/useProjectStore.ts
```diff
@@ -57,6 +75,10 @@ export const createNewProject = (projectType: ProjectType = DEFAULT_PROJECT_TYPE
   }
 })
 
+const nextIDFor = (elementArr: AutomataState[] | BaseAutomataTransition[] | ProjectComment[]): number => {
+  return 1 + Math.max(-1, ...elementArr.map((e: AutomataState | BaseAutomataTransition | ProjectComment) => e.id))
+}
```
ire4ever1190, line 80: We only need an `id` so we can just specify that (Also means it will work for any future items)
```suggestion
const nextIDFor = (elementArr: {id?: number}[]): number => {
  // We can't do elementArr.length + 1 since that might reuse an ID from a deleted item
  // Uses || -1 to handle any of the items being `undefined` (Shouldn't happen, but our type says it can)
  return 1 + (Math.max(-1, ...elementArr.map(e => e.id)) || -1)
}
```
I used `|| -1` since `Math.max(-1, undefined)` returns `NaN` (This shouldn't happen in practise though, except the type for `ProjectComment` needs to allow for undefined ID so this is more just defensive programming).

Also I can do this if you want, but `createState`, `createTransition`, and `createComment` should also be using `nextIDFor`

ire4ever1190 (changes requested):
Nice work, just a few comments to look over but otherwise looking really good 🚀

frontend/src/hooks/useActions.ts
SidhraFernando-Plant: Nah that's all good it makes sense, I think this was left over from when there were some different options of stuff to assign to the return

frontend/src/stores/useProjectStore.ts
SidhraFernando-Plant: I tried putting this function in, but it resulted in some unexpected behaviour: if I have an empty project, create a state (q0), and then copy paste it the next ID is not calculated correctly (it is pasted as q0 as well). 

So I was thinking, perhaps it could be left as is and this could be opened as a separate issue in the interest of prioritising getting the templates into dev? I'm sure it wouldn't be too hard to work out what the issue is and get this refactored version into the code, I am just keen to get this merged if possible. Perhaps calling the function in createState, createTransition, and createTransition could be tackled in that same issue? Let me know what you think

ire4ever1190 (approved):
Sorry about that. All looking good 👍 you can merge whenever

PR #349, opened 2023-04-21
Don't reorder when no initial state

ire4ever1190:
This disables the reordering action when there is no initial state instead of just treating `q0` as the initial and working from that.

It disables it in the frontend, and just does an early return in the function incase it is used elsewhere


LvandoApps (approved):
Good fix

PR #351, opened 2023-04-22
Refactor graph types

### LINKED ISSUE #301: Merge graph types together

ire4ever1190:
Closes #301 

This refactors the types used by the simulator to be a lot simpler and be more coherent with the rest of the project. 

The only change that might affect behaviour is that the `read` property is now a string for both simulator and frontend (This is automatically expanded when getting read from the store). This was done so that the types didn't need to change in the simulator

This also updates the types used with functions like `closureWithPredicate` and `validTransitions` so that the typing properly works for all graph types instead of just FSAs

The different transition types were separated out and then we use `projectType` has a discriminator to check which transitions we are allowed to use. This makes the autocomplete nicer since it will be correct for the current graph type that is getting used (If typescript can properly narrow the type, which in some cases it can't).

Also cleaned up the `State` types used by the simulator. The properties are now `readonly` instead of private with a getter

To handle the cases where typescript doesn't properly narrow types I added `assertType` which can assert the type of a variable inside a block of code. This is so you don't need to constantly add `var as SomeType` to everytime you use a variable. This is semi dangerous and should be used sparingly


frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.js
```diff
@@ -75,7 +65,7 @@ const TestingLab = () => {
             projectType === 'PDA'
               ? simulatePDA(graph, input ?? '')
               : simulateFSA(graph, input ?? '')
-
+      console.log(accepted, trace, remaining, graph)
```
LvandoApps, line 68: Ensure logs are removed

ire4ever1190:
Found one small problem, fixing now

LvandoApps (changes requested):
Looks good 👍, tests are updated correctly and types look much cleaner now. I don't understand absolutely everything that has been changed, however in terms of the type updating it makes sense and I am excited to see it incorporated into dev. Just do some basic cleanup of debug tools (such as console logs) and it'll be good to go, well done

LvandoApps (approved):

PR #352, opened 2023-04-22
Minimise number of trap states in DFA

LvandoApps:
Just a small fix so that there aren't an excessive number of trap states created. Trap states that are initially created by the user will remain as trap states (in case the user has done so for some specific reason), however in terms of trap states that are created by the algorithm during conversion, there should only be one that is used by all.

Old:
image

New:
Screenshot_2023-04-22_133504


packages/simulation/tests/convertNFAtoDFA.test.ts
```diff
@@ -156,4 +157,51 @@ describe('Check to ensure DFA graph is displayed as expected', () => {
...
+    expect(graph.states.length).toBe(4)
+    // Should be two final states, q2 and q3
+    expect((graph.states.find((state) => state.id === 0)).isFinal).toBe(false)
+    expect((graph.states.find((state) => state.id === 1)).isFinal).toBe(false)
+    expect((graph.states.find((state) => state.id === 2)).isFinal).toBe(false)
+    expect((graph.states.find((state) => state.id === 3)).isFinal).toBe(true)
+    // No lambda transition should be found, and should be 28 transitions total. All states should have a single transition 'A' to 'G' and be connected in a certain way
+    expect(graph.transitions.length).toBe(28)
+    expect(graph.transitions.some((transition) => transition.read === undefined)).toBe(false)
+    expect(graph.transitions.every((transition) => transition.read === 'A' || transition.read === 'B' || transition.read === 'C' || transition.read === 'D' || transition.read === 'E' || transition.read === 'F' || transition.read === 'G')).toBe(true)
```
ire4ever1190, line 174:
```suggestion
    const alphabet = 'ABCDEFG'
    expect(graph.transitions.every((transition) => alphabet.includes(transition.read))).toBe(true)
```
Bit redundant, you can just test both at once

packages/simulation/tests/convertNFAtoDFA.test.ts
```diff
@@ -156,4 +157,51 @@ describe('Check to ensure DFA graph is displayed as expected', () => {
...
+
+    // Final state with no transitions to must all go to a single trap state
+    expect(graph.transitions.filter((transition) => transition.from === 3).length).toBe(7)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'A')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'B')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'C')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'D')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'E')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'F')[0].to).toBe(2)
+    expect(graph.transitions.filter((transition) => transition.from === 3 && transition.read === 'G')[0].to).toBe(2)
```
ire4ever1190, line 205:
```suggestion
    const isTrapped = (from: number, toTrap: string) => {
      const transitions = graph.transitions.filter(t => t.from === from)
      expect(transitions.length).toBe(alphabet.length)
      // Find all the transitions that read a symbol we expect to go to a trap state
      const expectedTrapTransitions = transitions
        .filter(t => toTrap.includes(t.read))
        .map(t => t.to)
      expect(new Set(expectedTrapTransitions).size).toBe(1)
      // All trap transitions should be going to state '2' (The trap state)
      expect(expectedTrapTransitions[0]).toBe(2)
    }
    // Transitions that are not initially present should all go to a single trap state "2"
    isTrapped(0, 'CDEFG')
    isTrapped(1, 'AB')
    // This is the trap state, should go to itself for every symbol
    isTrapped(2, alphabet)
    // Final state with no transitions to must all go to a single trap state
    isTrapped(3, alphabet)
```

Try and group similar logic into functions. Espically for tests because if behavior changes in future then it is good if less spots need to be changed

ire4ever1190 (changes requested):
Tests are a bit messy, but least the trap states are being minimised :+1:

ire4ever1190 (approved):
Merge after #351 :+1:

LvandoApps:
This change ensures that bugs such as dib easteregg are removed, merging now

PR #354, opened 2023-04-25
Fix typescript linting

ire4ever1190:
Fixes the issue found in #348 where the linter considered a used enum to be unused.
Does this by using the recommended typescript rules which properly account for that.

This also meant that quite a few files needed to be updated because things that previously weren't caught are now caught (e.g. explicit `any` isn't allowed anymore)

The biggest change is is `utils.ts` where the `buildProblem` function I made needed to be cleaned up. The changes to it are a lot nicer since now the type inference gets the return type correct is most cases now. The clean up to the mapper types also means making generics for the simulator are a lot nicer

### DIFF
--- a/eslintConfig.json
+++ b/eslintConfig.json
@@ -3,7 +3,9 @@
     "es2021": true
   },
   "extends": [
-    "standard"
+    "standard",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended"
   ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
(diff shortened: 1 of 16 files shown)


LvandoApps (approved):
Looks good, good to see the TS takeover 🚀

ire4ever1190:
ew github ui doesn't let me change message

ire4ever1190:
Weird the tests didn't fail in your PR but I know the issue I think

ire4ever1190:
> > Weird the tests didn't fail in your PR but I know the issue I think
> 
> Hm strange, my guess is that the meta/config data in the DFA JSON could be decided in real time rather than by the details of the NFA JSON?

Dw, it's just cause we don't add the extra project data to the graph in the conversion because there are now actual functions in the projectStore for working with graph data

PR #355, opened 2023-04-25
Clean up presentation of convert tests

LvandoApps:
A small PR that just cleans up the presentation of the convert tests, nothing to do with functionality. In terms of the behavior of the transitions, since the requirements for them to be correct are extremely specific for each test, I found it difficult to automise them and have left them alone as a result, however everything else has been cleaned up appropriately.


packages/simulation/tests/convertNFAtoDFA.test.ts
ire4ever1190, line 1: Every test seems to just be doing object testing.
Might be easier to just store the expected DFA (probably in a seperate JSON file and just import) and then just compare directly.
Then adding a new test is a matter of creating both the NFA and expected DFA in the browser and just exporting the files

LvandoApps: Didn't even think of doing this, good idea, I'll swap to using isEqual for the two JSON's instead then

ire4ever1190 (approved):
Nice, lot cleaner :rocket:

PR #357, opened 2023-04-26
Refactor more to typescript

ire4ever1190:
image

Gets the percentage of Javascript down to 15% which is mainly a few files that aren't used much.

Was meant to be multiple small PRs but it got a it out of hand since I kept needing to convert other files to test files.

- Converts rest of the stores
- Converts most of the hooks. The few left require types from #342 
- Converted quite a few components and pages to typescript

Some handy notes for future conversions
- Make sure to inherit types correctly for components. Makes life a lot easier

One minorish change to the functionality of the program is that `halted` is now `accepted` in `TMExecutionResult` since having the different cases for acceptance was annoying


frontend/src/components/CommentRect/commentRectStyle.ts
```diff
@@ -12,7 +12,7 @@ export const commentStyles = {
   width: 'max-content',
   maxWidth: '255px',
   boxSizing: 'border-box'
-}
+} as const
```
ire4ever1190, line 15: Handy feature which makes the types become more restricted i.e. `boxSizing: 'border-box'` has type `'border-box'`, not `string`

frontend/src/hooks/useEgg.ts
ire4ever1190, line 1: Just like typescript, please ignore this file

ire4ever1190:
Few bugs I found that I'll fix quick 
- [ ] Rejected transition has wrong ending label

LvandoApps (approved):
Features are running as expected so should be good to go, nothing that breaks the matrix in any of the files from what I can see too, always appreciate a cheeky JS to TS merge

PR #364, opened 2023-05-01
Fix transitions not being formatted properly

ire4ever1190:
Issue was that the parameter wasn't being used so it was just using the first value for everything


SidhraFernando-Plant (approved):
Very speedy, thank you!

PR #365, opened 2023-05-01
Fix projects not being exportable

ire4ever1190:
Small bug that stopped exports. Issue was that it wasn't getting the actual value of the project


SidhraFernando-Plant (approved):
Solid 👌

PR #366, opened 2023-05-03
Fix not being able to move both states and comments

### LINKED ISSUE #214: Comments don't move when selected with states (and vice versa)

ire4ever1190:
Closes #214 

The issue was that the dragging for comments and states was handled separately and so you could only drag one or the other.

This also does conversion to typescript for the dragging related files


LvandoApps (approved):
Ah as in when you highlight them both? Interesting, good fix then, works well in the preview 👍

Issue #259, opened 2022-07-28
Reflect save state on save itself


tdib:
Saving animation currently occurs on the debounce between changing a project graph and the save itself. This animation should reflect what's actually happening, and only show when a project is being saved :)

PR #368, opened 2023-05-03
Check if we need to save, even if not logged in

### LINKED ISSUE #318: Saving doesn't work properly when project is exited too quickly
### Description of the bug

Saving does not work properly when you exit the project too quickly after making changes. The thumbnail updates correctly but changes disappear when you enter the file.

### To reproduce

1. Be in a project
2. Make a change
3. Exit the project quickly after making change
4. Thumbnail is updated, but last save date is not
5. Enter the project
6. Change is gone

### Expected behavior

Either saving should occur quicker, or it should be indicated to the user when saving is still in progress/prevent exiting project if project is not done saving.

### Additional information

_No response_

ire4ever1190:
Closes #318 

Problem was that the frontend only considered that it was trying to save if the user was signed in and there was a modification. This makes it only check modification time and ignore the user.

The check for user was moved into the saving icon since we don't want to show it if only showing locally and not syncing to a backend

### DIFF
--- a/frontend/src/components/Menubar/Menubar.tsx
+++ b/frontend/src/components/Menubar/Menubar.tsx
@@ -78,7 +78,7 @@ const Menubar = () => {
   const lastSaveDate = useProjectStore(s => s.lastSaveDate)
   const lastChangeDate = useProjectStore(s => s.lastChangeDate)
   // Determine whether saving
-  const isSaving = useMemo(() => user && !(!lastChangeDate || dayjs(lastSaveDate).isAfter(lastChangeDate)), [user, lastChangeDate, lastSaveDate])
+  const isSaving = useMemo(() => !(!lastChangeDate || dayjs(lastSaveDate).isAfter(lastChangeDate)), [lastChangeDate, lastSaveDate])
 
   const handleEditProjectName = () => {
     setTitleValue(projectName ?? '')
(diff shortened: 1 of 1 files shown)


frontend/src/components/Menubar/Menubar.tsx
```diff
@@ -136,7 +136,7 @@ const Menubar = () => {
                 : (
                 <Name onClick={handleEditProjectName} title="Edit title">{projectName ?? 'Untitled Project'}</Name>
                   )}
-              <SaveStatus $show={isSaving}>Saving...</SaveStatus>
+              <SaveStatus $show={isSaving && !!user}>Saving...</SaveStatus>
```
ire4ever1190, line 139: The `!!` is so its treated has a boolean.
It pleases the linter

SidhraFernando-Plant: I was just wondering why save status is not shown for not logged in users - is it because there's no point showing save status when someone isn't logged in because it happens so quickly it would be missed? Just wanted to double check I understand this correctly

SidhraFernando-Plant (approved):
Great to see this issue fixed, nice work 🎉

frontend/src/components/Menubar/Menubar.tsx
ire4ever1190: Yeah thats basically the reason, except the animation actually runs for a while which was a bit annoying imo (think this is what #259 is talking about)

PR #369, opened 2023-05-05
Remove initial state properly when it is deleted

SidhraFernando-Plant:
### Overview
The initialState property of the project store is now set to null when the initial state is deleted.

### Problem
The initialState property was not being reset if the initial state was deleted. Even after an initial state was deleted, if a state with the same ID was created it would be made initial.

### Solution
When a state is deleted, we check if it was an initial state. If it was, initialState is set back to null.


ire4ever1190 (approved):
Nice :+1:

PR #371, opened 2023-05-07
Fix not being able to drag items

ire4ever1190:
Issue was caused by the change to allow selecting a single item from a group selection.
This behavior is now changed to be a double click instead (i.e. if you select a whole group of states but then only want to select one, you double click it)

### DIFF
--- a/frontend/src/components/StateCircle/stateCircleStyle.ts
+++ b/frontend/src/components/StateCircle/stateCircleStyle.ts
@@ -18,4 +18,4 @@ export const circleSelectedClass = css`
 export const textStyles = {
   userSelect: 'none',
   fill: 'var(--stroke)'
-}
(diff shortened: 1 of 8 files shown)


frontend/src/components/StateCircle/stateCircleStyle.ts
```diff
@@ -18,4 +18,4 @@ export const circleSelectedClass = css`
 export const textStyles = {
   userSelect: 'none',
   fill: 'var(--stroke)'
-}
+} as const
```
ire4ever1190, line 21: Was getting type errors so it needed to be extra const so that the strings were restricted properly

SidhraFernando-Plant (approved):
Nice job sorting this out 👌🎉

PR #372, opened 2023-05-08
Fix the shortcuts page

ire4ever1190:
Fixes out of order hook calls and a weird access to make the shortcuts page (Hotkey: `ctrl + /`) work again.

Also makes all the hotkeys be arrays instead of some being `HotKey` and some being `HotKey[]` so that access is easier


LvandoApps (approved):
Much needed good fix

PR #373, opened 2023-05-08
Feat/warning modal

LvandoApps:
Adds a nice warning modal to be used in error messages instead of the disgusting dialog box provided by "alert". Open to suggestions as it is purely UX (text font is the big one).

### DIFF
--- a/frontend/src/hooks/useActions.ts
+++ b/frontend/src/hooks/useActions.ts
@@ -11,6 +11,7 @@ import { createNewProject } from '/src/stores/useProjectStore'
 import { reorderStates } from '@automatarium/simulation/src/reorder'
 import { convertNFAtoDFA } from '@automatarium/simulation/src/convert'
 import { FSAProjectGraph } from '/src/types/ProjectTypes'
+import { showWarningModal } from '/src/components/WarningModal/warningModal'
 
 /**
  * Combination of keys. Used to call an action
@@ -181,7 +182,7 @@ const useActions = (registerHotkeys = false) => {
         }
         let isInitialStateUpdated = false
         if (pasteData.projectType !== project.projectType) {
-          alert(`Error: you cannot paste elements from a ${pasteData.projectType} project into a ${project.projectType} project.`)
+          showWarningModal(`Error: you cannot paste elements from a ${pasteData.projectType} project into a ${project.projectType} project.`)
           return
         }
         const isNewProject = pasteData.projectSource !== project._id
@@ -215,7 +216,7 @@ const useActions = (registerHotkeys = false) => {
         })
         // Error if trying to paste transition without its to and from states
         if (newTransitions.find(transition => transition.from === null || transition.to === null)) {
-          alert('Sorry, there was an error while pasting')
+          showWarningModal('Sorry, there was an error while pasting')
           removeStates(pasteData.states.map(state => state.id))
           return
         }
@@ -321,7 +322,7 @@ const useActions = (registerHotkeys = false) => {
           updateGraph(reorderStates(convertNFAtoDFA(reorderStates(project as FSAProjectGraph))))
           commit()
         } catch (error) {
-          alert(error.message)
+          showWarningModal(error.message)
         }
       }
     },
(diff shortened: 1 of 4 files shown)


SidhraFernando-Plant (changes requested):
This is a good start to the modal and very speedily done 😲 the positioning of the modal is good and it clearly looks 'in front' of the page which is great. I have a few design suggestions:
Screen Shot 2023-05-08 at 3 20 29 pm
- Decrease the width of the modal
- Make the button take up the full width of the modal (shown in image), or if that doesn't look good just make the button wider
- Decrease spacing between heading and body text (shown in image)
- Remove outline, the shadow is enough to distinguish it from the background
- An icon above the heading would be nice to have, but not essential. The rest of the project uses lucide icons, the alert-triangle or x-octagon could be good
- This is a nitpick, but I find 'Warning' to not necessarily match all use cases for the modal, particularly how I have used it in copy paste code as it represents an error there as opposed warning the user about something. I think 'Alert' could be better, or the heading could even be passed as a parameter and customised when the modal is called.
- Use the CSS colour variables defined in style.css to ensure consistency with the rest of the project (ie. the white black and error variables particularly)

Happy to discuss any of these suggestions

tdib:
This was defs a needed improvement, and just to get my 2 cents in, I would say try to stick to the design of other modals that already exist (e.g. below)

image

That way we get 1) consistency across components, and 2) it reduces your workload by being able to yoink styles and such from other components :)

In regards to Sidi's comment, I think making a full width button would look a bit strange, and that the modal would generally look better if the stuff was left aligned - I think it's easy to fall into a trap of centering everything but it ends up looking a bit odd.

Nice work anyways, keep it up 😎

SidhraFernando-Plant:
Yeah for sure that is a great idea, I had forgotten that there were already modals in the application. I will check it out now :)

LvandoApps:
No worries I did as well hahaha, good thing we have our elders to lead us the right way 🙏

frontend/src/components/Warning/Warning.tsx
```diff
@@ -0,0 +1,27 @@
+import { useState } from 'react'
+import { useEvent } from '/src/hooks'
+import { Modal, Button } from '/src/components'
+
+export const Warning = () => {
+  const [isOpen, setIsOpen] = useState(false)
+  const [message, setMessage] = useState('')
```
SidhraFernando-Plant, line 7: nit: Perhaps message does not need to use state? I think it would be fine to pass it as a prop as it won't change while the modal is rendered

SidhraFernando-Plant: Nevermind I was misunderstanding how this works let me take a closer look at this

SidhraFernando-Plant (dismissed):
This is looking great and is definitely more cohesive with the project. Just one small thing I think could be improved is to reduce the spacing between the warning heading and the body message. Otherwise, looking very nice 👍 
Screen Shot 2023-05-08 at 10 10 32 pm
Something like this I reckon (I just quickly sketched this out in dev tools) a bit more space would be okay this is just a rough visual

LvandoApps:
Good suggestion, the modal already came with a feature that lays it out pretty much that way so I've switched over to using that instead. Take a look and let me know what you think.

frontend/src/hooks/useActions.ts
```diff
@@ -590,4 +590,8 @@ const promptLoadFile = (parse, onData, errorMessage = 'Failed to parse file') =>
   input.click()
 }
 
+function showWarning (msg: string) {
+  dispatchCustomEvent('showWarning', msg)
+}
+
```
ire4ever1190, line 596: Useful function
Put it inside the warning file and just import it in other places

Maybe a cheeky little bit of documentation saying it shows the modal 👀

ire4ever1190 (changes requested):
Looking snazzy 👌 just one small thing

SidhraFernando-Plant:
Yep that LGTM now. Nice job! Will leave Jake to approve once his changes are in

frontend/src/hooks/useActions.ts
LvandoApps: Good suggestion, all done

ire4ever1190 (approved):
epic 🚀

PR #374, opened 2023-05-08
Only accept `.jff` and `.json` files in file importer

ire4ever1190:
Noticed in #373 that the file importer allows for any files which means the user has higher chance of clicking the wrong file and that every file is shown instead of the ones that the user is interested in.

This restricts the file input to only accept `.jff` and `.json` files (this is case insensitive, so `.JFF` is also accepted)

I'm leaving the file extension check in since the user can still override the browser and select anything


SidhraFernando-Plant (approved):
Love this UX enhancement 😎

PR #370, opened 2023-05-06
Better multiselecting/unselecting

ire4ever1190:
Closes #262 

Fixes a few problems with multiselection/unselection
- You can now unselect a single item from a multi selection by holding shift and clicking it
- Holding shift and selecting a group of items now appends that group to the current selection
- Clicking a single item (without holding shift) in a multiselection now only selects that single item

PR #376, opened 2023-05-09
Fix regression for right clicking a state

### LINKED ISSUE #375: State no longer selected on right click
### Description of the bug

When right clicking on a state, it does not select the state like it used to.

This results in all actions on the right click menu (right click on a state specifically) doing nothing when clicked as they do not work without a state selected.

This issue was discovered when a user was trying to set a final state in a test and was right clicking on the state to toggle is final without selecting the state, so it didn't work and even I didn't immediately realise why it wasn't working.

Screen Shot 2023-05-09 at 5 45 05 pm

### To reproduce

1. Right click on a state
2. See options not be disabled
3. Click on a not disabled option, nothing happens

### Expected behavior

Right clicking on a state should select the state automatically. We will have to be mindful that the case of selecting a group of states, and then right clicking on one to apply an action to ALL selected states is not broken ie. the right click on a state should not affect a group that is already selected by only selecting that state instead.

### Additional information

_No response_

ire4ever1190:
Closes #375 

Fix issue I caused in #370 that stopped right clicking from selecting. I had a mental note to move the code outside the if statement but kept forgetting to :stuck_out_tongue_closed_eyes:

I also made final state selection be disabled like initial state. This doesn't really change anything, but just makes it clearer that something needs to be selected

### DIFF
--- a/frontend/src/components/EditorPanel/EditorPanel.tsx
+++ b/frontend/src/components/EditorPanel/EditorPanel.tsx
@@ -37,12 +37,12 @@ const EditorPanel = () => {
   useContextMenus()
 
   const handleDragging = (e: SelectionEvent) => {
+    // Only try and check if the user is selecting a new resource if the event correlates with that.
+    // Else just use the previous value from the store.
+    // Outside the if so that you can directly right-click and edit a state
+    const selStates = e.type === 'state:mousedown' ? selectState(e) : selectedStates
+    const selComments = e.type === 'comment:mousedown' ? selectComment(e) : selectedComments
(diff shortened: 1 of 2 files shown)


SidhraFernando-Plant (changes requested):
With the disabling final state part, it is disabling it when multiple states are selected as well so now user cannot set all states in a selection to be final states.

Otherwise looks good though, thank you for the speedy fix 🔥

ire4ever1190:
Good catch :ok_hand: handy feature, didn't know it existed

SidhraFernando-Plant (approved):
Nice job 🚀

Issue #379, opened 2023-05-11
Kebab menu


SidhraFernando-Plant:
### Proposed feature or solution

At the moment the kebab menu on the ProjectCard component doesn't do anything. There is a delete function tied to the menu that is not functional. Either remove the icon, or make it functional.

ire4ever1190:
Related #231

SidhraFernando-Plant:
The menu in question
![Screen Shot 2023-05-11 at 1 07 24 pm](https://github.com/automatarium/automatarium/assets/79517325/d90f1045-db59-4bef-9295-d40d50bfbbcb)

Console when you press it
![Screen Shot 2023-05-11 at 1 08 23 pm](https://github.com/automatarium/automatarium/assets/79517325/091f25ca-7fce-4526-a422-bfa21fa9679f)

Issue #231, opened 2022-06-25
Delete projects


GRA0007:
Projects should be able to be deleted by the user

PR #381, opened 2023-05-11
Templates Part 2 - UI

SidhraFernando-Plant:
### Overview
Implement the UI for the templates feature.

Branch train:
1. #348 
2. #381 ⬅️ this PR

This PR will be merged into #348 which will then be merged into dev.

### Features
Templates can now be:
- Added with a name chosen by the user
- Inserted, with a 'ghost' version of the template displaying on mousedown
- Deleted, by pressing the backspace key when a template is selected and nothing else in the project is selected

Additional features:
- A user's templates are displayed with the most recently created showing first. Only templates that match the current project type will be shown.
- Once a user selects a template, they can insert it as many times as they want. To stop inserting the template, they can either click on the template again to deselect, or close the templates side panel, or select a tool.

Template thumbnails have not yet been implemented but will be added before merging this PR.

### Future improvements
- Improve UI for deleting a template, this goes hand in hand with the deleting of projects as the templates use the same ProjectCard components (related to issues #379 and #231 )
- Add tooltips to the Templates sidepanel
- Add info about templates code to project documentation


frontend/src/hooks/useActions.ts
```diff
@@ -193,14 +179,20 @@ const useActions = (registerHotkeys = false) => {
...
+        if (template !== null && selectedCommentsIds.length === 0 && selectedStatesIds.length === 0 && selectedTransitionsIds.length === 0) {
+          if (window.confirm(`Are you sure you want to delete your template '${template.name}'?`)) {
+            deleteTemplate(template._id)
+            setTemplate(null)
+            setTool('cursor')
+            return
+          }
+        }
+        // Otherwise, delete selection
+        if (selectedStatesIds.length > 0 || selectedTransitionsIds.length > 0 || selectedCommentsIds.length > 0) {
```
ire4ever1190, line 192:
```suggestion
        else if (selectedStatesIds.length > 0 || selectedTransitionsIds.length > 0 || selectedCommentsIds.length > 0) {
        // Otherwise, delete selection 
```
imo this is better since it shows that only one branch can run

frontend/src/hooks/useCreateBatch.ts
```diff
@@ -1,3 +1,5 @@
+// NOTE: THIS HOOK IS NOT CURRENTLY USED/CALLED ANYWHERE
+// SEE useProjectStore FOR EQUIVALENT FUNCTIONALITY THAT IS ACTUALLY CALLED
```
ire4ever1190, line 2: Is this going to be used in the original pr or something?

frontend/src/components/ProjectCard/projectCardStyle.ts
```diff
@@ -11,7 +11,7 @@ export const CardContainer = styled('button')`
   color: inherit;
   cursor: pointer;
 
-  &:hover, &:focus {
+  &:hover${props => !(props.istemplate === 'true') && ', &:focus'} {
```
ire4ever1190, line 14: Was there an issue with a boolean?

ire4ever1190 (changes requested):
Looking good :rocket:, just a few minor things

frontend/src/components/ProjectCard/projectCardStyle.ts
SidhraFernando-Plant: Yes 😢 This is what I got when trying to pass in istemplate={true}
Screen Shot 2023-05-29 at 1 53 23 pm

ire4ever1190: Ah I know the problem

Put `$` before the name so it becomes `$isTemplate`. Goober is setup to not pass properties that are prefixed with `$` to the DOM so we can use them for pure logic (least I think thats what it does)

tdib: Memory is a bit hazy but I thought Goober was set up to _prefer_ transitive ($) properties? As always the expert is Benji or Ewan :)

ire4ever1190:
```ts
  // Remove transient props from the DOM
  props => Object.keys(props).forEach(p => p[0] === '$' && delete props[p])
```
Seems to remove it, although I'm unsure in my answer now since I can't seem to reproduce the warning? So don't know if doing that will actually fix it

ire4ever1190: Ok was able to finally reproduce :rocket:
Suggestion works, changing to `$isTemplate` and updating `Templates.tsx` to use the new property (which I wasn't doing when checking which is why I got no error) allows to use booleans instead

SidhraFernando-Plant: Thanks for getting to the bottom of this! Will update it

frontend/src/hooks/useCreateBatch.ts
SidhraFernando-Plant: In the original PR i kind of left it there as a reference for how I had tried to do this stuff as a hook (but I couldn't call it properly). I had left it in there in case reviews indicated it should be used if how I had done it in the store wasn't suitable. Kind of like 'this was an alternative solution attempt that didn't pan out but could be brought back if necessary'. But seeing as there hasn't been any comments indicating to use the hook instead of what I did in the store I will just delete it.

SidhraFernando-Plant:
Re-requesting review, will discuss plans for merging/thumbnails in meeting today

ire4ever1190 (approved):
Nice, looks good to me 🚀

frontend/src/components/Sidepanel/Panels/Templates/Templates.tsx
```diff
@@ -1,27 +1,114 @@
...
+import React, { useState } from 'react'
+import { Template } from '/src/types/ProjectTypes'
+import { TEMPLATE_THUMBNAIL_WIDTH } from '/src/config/rendering'
+import { WarningLabel } from '../TestingLab/testingLabStyle'
+import { Tool } from '/src/stores/useToolStore'
+
+export const stopTemplateInsert = (setTemplate: (template: Template) => void, setTool: (tool: Tool) => void) => {
+  setTemplate(null)
+  setTool('cursor')
+}
```
SidhraFernando-Plant, line 19: Put this into its own function, as it's used a few times across the project

frontend/src/components/Sidepanel/Sidepanel.tsx
```diff
@@ -41,7 +44,7 @@ const panels: PanelItem[] = [
     element: <Options />
   },
   {
-    label: 'Templates',
+    label: 'Templates (Beta)',
```
SidhraFernando-Plant, line 47: Add Beta label

frontend/src/components/Sidepanel/Sidepanel.tsx
```diff
@@ -77,7 +85,10 @@ const Sidepanel = () => {
         {panels.map(panel => (
           <SidebarButton
             key={panel.value}
-            onClick={() => setActivePanel(activePanel?.value === panel.value ? undefined : panel)}
+            onClick={() => {
+              setActivePanel(activePanel?.value === panel.value ? undefined : panel)
+              stopTemplateInsert(setTemplate, setTool)
+            }}
```
SidhraFernando-Plant, line 91: Make it so that template insert mode stops when you switch to another sidepanel, before it was only when you close side panel

SidhraFernando-Plant:
Just made a few small changes which I've highlighted with comments

ire4ever1190 (approved):
Latest changes looking good :rocket:

PR #384, opened 2023-05-13
Convert last files to typescript

### LINKED ISSUE #277: Migrate to typescript
Currently the backend and some of the packages are written in typescript. It would be beneficial to move the frontend to typescript so that everything is using the one language and reduce the number of runtime bugs we could encounter

ire4ever1190:
image

Closes #277 

This converts the last files to typescript. Most of the remaining files were auth related (signing in, creating account, etc) and since the backend isn't set up, I had know way to test it. Although there wasn't many changes so unlikely to have problems


LvandoApps (approved):
Nice!

PR #385, opened 2023-05-15
Remove the warnings from the console

### LINKED ISSUE #367: Reduce warnings in console

Currently there are a lot of warnings in the console which makes debugging a bit annoying. This isn't that big of a deal, more of an annoyance

image

### Alternatives

_No response_

### Additional information

_No response_

ire4ever1190:
Closes #367 

The issue was some attribute where getting incorrectly used

- Width and height needed to passed as style properties, the `width` and `height` attribute [are for pixels](https://www.w3schools.com/tags/att_width.asp) and so `rem` cannot be used. This caused some uses of the logo to become smaller since it was now using the proper size, I made these instances use a blank size which achieves the same effect
- `markerUnits` only accepts [userSpaceOnUse and strokeWidth](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/markerUnits). I just used `strokeWidth` since it is used in other places and this didn't change any looks for me

### DIFF
--- a/frontend/src/components/Logo/Logo.tsx
+++ b/frontend/src/components/Logo/Logo.tsx
@@ -18,7 +18,10 @@ const Logo = ({
 }: LogoProps) => (
   <svg
     xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"
-    width={size} height={size}
+    style={{
+      width: size,
+      height: size
+    }}
     {...props}
   >
     {!hidePlanet && <circle fill={secondary} cx="241.33" cy="241.33" r="173.99"/>}
--- a/frontend/src/components/ProjectCard/ProjectCard.js
+++ b/frontend/src/components/ProjectCard/ProjectCard.js
@@ -17,7 +17,7 @@ const ProjectCard = ({ name, type, date, image, projectId, ...props }) => {
   const deleteProjectFromStore = useProjectsStore(s => s.deleteProject)
   return <CardContainer {...props}>
     <CardImage $image={!!image}>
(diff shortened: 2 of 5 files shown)


SidhraFernando-Plant (commented):
This is looking good, relieved to get rid of those 😅

I've found another two while checking out your changes that I hadn't encountered before:

I got this one under very specific conditions, I was in a project, then I clicked the title to start editing it, then exited while the title was still a text input (this was an accident)

Screen Shot 2023-05-15 at 2 30 44 pm

Then I looked more into image export as I noticed the previous error was linked to it and found this when I tried to export the project as an image:

Screen Shot 2023-05-15 at 2 32 38 pm

I don't mind if we open a separate PR for these or look at it in this one, if we wanna do a separate PR that's okay just let me know and I'll approve this one

SidhraFernando-Plant:
I should also note that the image export didn't work along with the console error

ire4ever1190:
Weird that those are popping up? Will see if they are cause of this PR

image exporting might be unrelated, there are quite a few existing problems with

ire4ever1190:
Both issues exist on dev, will fix in a separate PR 

Thanks for finding them :ok_hand:

SidhraFernando-Plant:
Oh yeah I'm pretty sure they're not related to this PR. So it would be fine if we tackle them as a separate PR I just put them here bc they are also console warnings (although as functionality is also impacted this doesn't really make sense whoops it ended up being more significant than I initially thought)

SidhraFernando-Plant (approved):

PR #389, opened 2023-05-26
Pass innerRef directly to remove react error

ire4ever1190:
I had this change but seems I forgot to commit it in the PR.

This just tells React that it should expect the `innerRef` parameter so it doesn't error in the console anymore (didn't seem to affect functionality, but was annoying)


LvandoApps (approved):
👍

PR #390, opened 2023-05-26
Disable copying JPEG images

### LINKED ISSUE #264: JPEG can't copy to clipboard
Attempting to export and then copy a JPEG to clipboard will throw the following error:
image
PNG copy to clipboard works fine.

ire4ever1190:
Closes #264 

Currently there is no support for [image/jpeg](https://stackoverflow.com/a/69536762) MIME type and the [custom formats spec](https://chromestatus.com/feature/5649558757441536) is still currently under development. I tried other methods like using the [copy command](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#copy) so that it would work on both chrome and firefox out of the box, but this had issues where pasting it would paste the base64 encoded image.

So instead I just disabled copying JPEGs for now, if chrome ever adds support then it should be easy to add back. If the user does want to copy they can just right click the image and hit copy (which does make me wonder if the "Copy to clipboard" is even needed for images (it still has use with SVGs)).

Sorry for the unrelated changes, blame the formatter not me 😛

### DIFF
--- a/frontend/src/pages/ExportImage/ExportImage.tsx
+++ b/frontend/src/pages/ExportImage/ExportImage.tsx
@@ -72,12 +72,12 @@ const ExportImage = () => {
     if (type === 'svg') {
       // TODO: kinda broken, might need to clean the svg at an earlier stage
       const blob = new Blob([svg], { type: 'text/plain' })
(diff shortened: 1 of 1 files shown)


LvandoApps (approved):
Will have to bring this up in the report for sure, see what future groups can do something with it

PR #391, opened 2023-05-29
Better types

ire4ever1190:
Cleans up some of the types and also introduces types for certain sections that were missing them. Mainly just typing changes so doesn't affect actual app.

Used the `noImplicitAny` option in `tsconfig.json` to find the spots that were lacking. Can't enable it by default though since there were some hairy sections that couldn't be fixed


LvandoApps (approved):
Yep looks good

PR #392, opened 2023-05-30
Use `nextIDFor` in places that need a new ID

ire4ever1190:
Followup to a comment in #348, this uses the `nextIDFor` in all the places that we calculate the ID.
Issue with my original comment was that if the highest was 0 then `0 || -1` would evaluate to -1 and so the ID returned would be 0 (when it should be 1).

I instead updated the type on `ProjectComment` to reflect that the `id` isn't optional since its given one when created and then did away with my original `NaN` check

### DIFF
--- a/frontend/src/stores/useProjectStore.ts
+++ b/frontend/src/stores/useProjectStore.ts
@@ -80,13 +80,10 @@ export const createNewProject = (projectType: ProjectType = DEFAULT_PROJECT_TYPE
  * but instead returns the number of the highest
  * e.g. [1, 4, 7] next ID would be 8
  */
-const nextIDFor = (elementArr: {id?: number}[]): number => {
+const nextIDFor = (elementArr: {id: number}[]): number => {
   // We can't do elementArr.length + 1 since that might reuse an ID from a deleted item.
-  // Order also is guarenteed so we can't just get the last element.
-  let highest = Math.max(-1, ...elementArr.map(e => e.id))
-  // If any of the IDs were undefined then `highest` would be NaN
-  if (isNaN(highest)) highest = -1
-  return 1 + highest
+  // Order also is guaranteed so we can't just get the last element.
+  return 1 + Math.max(-1, ...elementArr.map(e => e.id))
 }
 
 interface ProjectStore {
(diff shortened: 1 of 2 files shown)


frontend/src/stores/useProjectStore.ts
```diff
@@ -212,8 +219,8 @@ const useProjectStore = create<ProjectStore>()(persist((set: SetState<ProjectSto
   })),
 
   /* Create a new comment */
-  createComment: (comment: ProjectComment) => {
```
ire4ever1190, line 215: I updated the type in the interface and with type inference we don't need the type here also

SidhraFernando-Plant: Hmm, I noticed that doing the same thing for `createState` (removing explicit type of `AutomataState`) results in an error if the inferred type of `Omit<AutomataState, 'isFinal' | 'id'>` is used, seeing as the `isFinal` attribute is optional as opposed to omitted and it used in the function body. Perhaps this typing should be changed for consistency? The typing for all of the comment stuff is good though

frontend/src/stores/useProjectStore.ts
```diff
@@ -99,7 +106,7 @@ interface ProjectStore {
   setName: (name: string) => void,
   createTransition: (transition: BaseAutomataTransition) => number,
   editTransition: (transition: Omit<BaseAutomataTransition, 'from' | 'to'>) => void,
-  createComment: (comment: ProjectComment) => number,
+  createComment: (comment: Omit<ProjectComment, 'id'>) => number,
```
SidhraFernando-Plant, line 109: Is it worth doing something like this for `createTransition` as well? As opposed to casting the parameter when it is called (in `frontend/src/hooks/useTransitionCreation.ts`), then it will be consistent with the other functions and I also find the Omit typing to be nice and clear as to what is missing (so nice job with that 😁)

SidhraFernando-Plant (commented):
All of the new nextIDFor/projectComment code is looking good, I have just commented with a few things I noticed from existing code that I think would be good to update while looking at the various create functions. If any are too significant they could be brought out into a separate PR

frontend/src/stores/useProjectStore.ts
ire4ever1190: Think I did it locally but forgot to commit 🤔 either way it's good idea so will do 🫡

frontend/src/stores/useProjectStore.ts
ire4ever1190: Will look into it 👍

SidhraFernando-Plant (approved):
Looks great! Thanks for refactoring this 🙏

PR #393, opened 2023-06-06
Fix reordering sort and small cleanup

ire4ever1190:
Was looking over the implementation and noticed I did sorting wrong (e.g. it would think `10 < 4`) so added test case and then fixed that up. 

Did a bit of clean up while I was there
- Use `t` when iterating through transitions like is convention with other places
- Clean up logic inside frontier exploration. Removes redundant checks and saves filtering the list twice
- Use optional access instead of creating a temporary list
- Use `Map` instead of `Object` since it makes more sense and is usually more efficient (not that it really matters at this small a scale)


LvandoApps (approved):
10 is certainly not less than 4, good catch lil bro

PR #394, opened 2023-06-09
Move graph reordering from right-click menu into the Tools dropdown.

ire4ever1190:
I felt it makes sense to have it in Tools since it doesn't depend on where the user right clicks

image


LvandoApps (approved):
Yeah was thinking this too, good change

PR #399, opened 2023-06-13
Refactor NFA -> DFA conversion

### LINKED ISSUE #397: NFA to DFA same symbol transition from one state with one being a self transition
### Description of the bug

To my knowledge the NFA to DFA converter works perfectly except under one very specific circumstance, which is when the machine has a state where there are two or more transitions going from it with the same symbol, and one of those transitions is to itself. In this case the machine will break as it will enter an infinite loop. This is because the way the algorithm works is that when it encounters a situation where there are multiple transitions from one state with the same symbol, it will commence a sort of "merge" where it merges the "to" states together and have them be "from" the original state, resulting in just one transition for the symbol. In the case where the original state is considered in this merge, there will be obvious problems. 

### To reproduce

Screenshot 2023-06-13 142243

### Expected behavior

It should merge all the states normally into their own states but EXCLUDE the original state (the one with the self transition), which will lead to the original state having a transition to the new merged state with the given symbol, but no self transition. This self transition is to then be added to the new merged state, and for any same symbol transitions from one state the process will continue until this is no longer the case.

Screenshot 2023-06-13 142837

### Additional information

I will work over the break to try and fix this issue, however any issue with this algorithm in the future should it come up, please inform me and I will either a) Fix it or b) Explain the algorithm if you don't understand it from the comments I laid out and want to fix it yourself.

ire4ever1190:
Closes #397

Spoke to Lachie and we decided it would be good to refactor `convert.ts`. This also comes with the added bonus of having labels like JFLAP does

**NFA**
image

**DFA**
image

All the tests are passing along with the extra test for #397


LvandoApps (approved):
Very clean, coming in with the powerful new features made even more powerful 🤩

PR #400, opened 2023-06-13
Batch updates during dragging for increased performance

ire4ever1190:
Performance would sometimes be sluggish on my laptop or with large machines so I profiled the code to find the problem.
Issue is that updating state is a very expensive operation for react and dragging would update the state a lot. By batching the updates together this improves the performance by a lot.

Another thing I tried to improve performance was move everything that used `number[]` to use `Set<number>` instead but that didn't help with performance

**Before PR**
slow

**After PR**
fast

[Benchmark file used](https://pastebin.com/75AuYB2c)


LvandoApps (approved):
Interesting that the performance change was this drastic, looks good though

PR #401, opened 2023-06-13
Only accept valid files when importing inside a project

ire4ever1190:
Same thing in #374, thought they used the same dialog but turns out they use different ones.

Would be a good idea for the next group to merge them into one


LvandoApps (approved):
Good to clarify

PR #402, opened 2023-07-01
Fix items being dragged even if they are not selected

ire4ever1190:
Hopefully this is the last bug squashed from #370. Issue was states that were previously selected were still being dragged even if not selected.

**Before**
Dragging being spooky

**After**
Recording 2023-07-01 at 22 22 30


frontend/src/hooks/useTransitionCreation.ts
ire4ever1190, line 1: Changes here are just typing related.
Not really relevant, but I noticed that the types were `any` when debugging so decided to clean it up

LvandoApps (commented):
Looks good 👍

LvandoApps (approved):

PR #418, opened 2023-08-14
Transition edge context menu

### LINKED ISSUE #398: Right click on transition arrow not working
### Description of the bug

There is a transition-specific right click menu, it currently only comes up if you right click on the label of the transition. It won't come up if you right click the arrow itself.
Screen Shot 2023-06-13 at 7 39 45 pm

### To reproduce

1. Be in project of any type
2. Right click on transition (arrow part specifically)
3. Transition edit/delete menu doesn't come up

### Expected behavior

The transition right click menu (with edit and delete options) should also come up when the arrow part of the transition is clicked (as well as when the label is clicked which is what currently happens, this should not be changed).

### Additional information

_No response_

anioncat:
Closes #398.

I was wondering what actions right-clicking an edge would give since clicking on a transition and clicking on an edge has slightly different meanings. I decided on a mass dialog box containing all transitions on the edge. This means even if only 3/6 transitions were selected, the dialog will display the full 6 transitions in the dialog, as the "edge" should include all these transitions.

20230814-182058_firefoxDubwq87K_540x305

20230814-182106_firefoxVFZ8x8ya_1094x667

Editing any of the inputs in the dialog will be updated in the project state, as well as adding new transitions, so this also serves as one of the ways to add many transitions quicker than dragging each one.

On the undo history: I have made it commit when the dialog is closed, so undoing will undo the entire edit from this dialog as opposed to undoing every single change in the dialogs. This was done to not clog up the undo history but I'm not 100% sure which is better from a usability perspective.

Other notes:
- Up/down arrow keys to move up and down the list
- Enter on the transition list will move focus down
- Enter on the new transition row will create a new transition
- The X on each row will delete the transition

Future todos:
- There are a lot of similarities to InputDialogs, refactoring would be nice
- TM input is validated but it may not be clear what the rules are to the user (single character, [LSR])
- Make look nicer?

### DIFF
--- a/frontend/src/components/ContextMenus/ContextMenus.tsx
+++ b/frontend/src/components/ContextMenus/ContextMenus.tsx
@@ -7,6 +7,7 @@ import graphContextItems from './graphContextItems'
 import stateContextItems from './stateContextItems'
 import transitionContextItems from './transitionContextItems'
 import commentContextItems from './commentContextItems'
+import edgeContextItems from './edgeContextItems'
 import { ContextItems } from './contextItem'
 
 const ContextMenus = () => {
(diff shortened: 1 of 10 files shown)


S3865562 (approved):
Works really well, nice one :)

PR #419, opened 2023-08-14
Kebab button

Christopher-Truong-s3848927:
Made the Kebab Buttons on project cards functional. Links don't do anything yet but can be easily be made to delete, rename etc.

### DIFF
--- a/frontend/src/components/KebabMenu/KebabMenu.tsx
+++ b/frontend/src/components/KebabMenu/KebabMenu.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react'
+import Dropdown from '../Dropdown/Dropdown'
 
 const KebabMenu = (props) => {
   const [open, setOpen] = useState(false)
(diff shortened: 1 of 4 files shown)


frontend/src/components/KebabMenu/KebabMenu.tsx
anioncat, line 1: Could I suggest using the already existing Dropdown component? Something like this:

```tsx
return <div>
      <a onClick={(e) => handleclick(e)}>
        {props.icon}
      </a>
      <Dropdown
        visible={open}
        onClose={() => setOpen(false)}
        style={{
          position: 'absolute',
          background: 'hsla(var(--primary-h), var(--primary-s), var(--primary-l), 0.9)'
        }}
        items={kebabItems}
      />
    </div>
```

Where `kebabItems` is something like
```ts
const kebabItems = [
  { label: 'Label1', action: actionFromContext },
  ...
]
```

frontend/src/components/ProjectCard/projectCardStyle.ts
```diff
@@ -67,6 +67,20 @@ export const TypeBadge = styled('div')`
 
 export const CardDetail = styled('div')`
   margin-block-start: 1em;
+  list-style-type: none;
+
+  .kebab-menu-options {
```
anioncat, line 72: Suggest using goober instead of html classes.

```ts
export const KebabMenuOptions = styled('div')`
  style options
`
```

Then you can use:
```tsx
<KebabMenuOptions></KebabMenuOptions>
```
in the react code.

anioncat (changes requested):

anioncat (approved):
lgtm.

I'll probably move around the context item stuff when I implement the delete project into this once it's merged, but it works for now.

PR #421, opened 2023-08-22
Using comment tool on non canvas objects

### LINKED ISSUE #245: Comments aren't created when clicking on states or transitions
When the comment tool is selected, a comment can only be created by clicking an empty part of the canvas. Clicking on an existing state or transition causes nothing to happen.

anioncat:
Closes #245.

Looked through the issue tracker and thought this was a quick fix.

- Added listeners on the other objects
- Removed the hit SVG boolean from the mouse event check

20230822-205150_firefoxtAxcA5kt_632x259


S3865562 (approved):
Looks good :)

Issue #414, opened 2023-08-08
Stepping lab - glow does not go away + colour


SidhraFernando-Plant:
### Description of the bug

When using the stepping lab, a glow appears around states to highlight the steps. Once this glow appears, it never goes away - it remains after closing the stepping lab and even after opening different projects or in a new project. For example if glow appears around `q1` while using the stepping lab, when you open a different project state `q1` (if present) will still have a glow. Additionally, I can't figure out how to get rid of the glow even in the same project - is it meant to disappear when the stepping lab is closed, or is there another intended way of making it disappear? Demo below:

https://github.com/automatarium/automatarium/assets/79517325/23c70a05-82bb-4244-971f-96158f699176

### To reproduce

1. Be in project with automata that can use the stepping lab (ie. initial and final states, transitions)
2. Use the stepping lab to achieve a glow around one of the states (let's say a glow is shown around state `q1`)
3. Exit the stepping lab, glow is still there
4. Go to another project
5. If there is a state with the same id (ie. `q1`) it will still be glowing
6. If there is no state with the same id, create such a state and it will have a glow

### Expected behavior

When the user closes the stepping lab the glow should disappear (a 'clear' button would also work, could use the X button that is already in the stepping lab to clear the glow/current stepping process). The glow should also disappear when the user navigates to a different project. 

### Additional information

This bug was discovered in user testing, where the testers took a while to notice the glow when using the stepping lab. They suggested using a more prominent colour for the glow so that it is more easily noticed, or adding some kind of prompt that indicates where they should be looking when using the stepping lab. This is something to also consider when working on this issue.

PR #422, opened 2023-09-02
Fix stepping lab functionality issues and appearance

### LINKED ISSUE #413: Stepping lab not functioning correctly
### Description of the bug

When I use the stepping lab, it will not properly step through states (getting stuck after the first step). This behaviour was initially observed in user testing.

Demo below:

https://github.com/automatarium/automatarium/assets/79517325/d63e3753-5c62-44c8-a418-d302869e8370

### To reproduce

1. Create FSA with initial and final states with at least 3 states (to allow more than one step)
2. Go to Stepping Lab
3. Input a string that would be accepted by the FSA into the Stepping Lab
4. Try to step through the string - it stops moving forward after one step (can still go backwards)

### Expected behavior

The stepping lab should continue to step through the string/states rather than stopping at the first one.

### Additional information

_No response_

S3865562:
Closes #413 and #414 

Changes:

- Removed the usage of setFrontier to avoid unwanted React state refreshes that were causing stepped states to reload
- Removed appearance of the glow circle when the Stepping Lab is not the active panel
- Increased the glow circle's intensity and radius to enhance visibility
- Removed ability to forward step when stepper returns non-empty frontier

Possible future enhancements:

- Disable buttons when movement is not possible
- Provide a visual indicator for the character being used in the traversal
- Add an alert or visual cue when the trace input is not accepted by the machine

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/SteppingLab/SteppingLab.tsx
+++ b/frontend/src/components/Sidepanel/Panels/SteppingLab/SteppingLab.tsx
@@ -11,7 +11,7 @@ import {
 } from 'lucide-react'
 
 import { graphStepper, Node, State } from '@automatarium/simulation'
-import { useMemo, useState } from 'react'
+import { useMemo } from 'react'
 import { FSAProjectGraph, PDAProjectGraph } from '/src/types/ProjectTypes'
 
 // The initial states and project store calls, along with the display of the current input trace,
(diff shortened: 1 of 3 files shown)


anioncat (approved):
Looks good.

Additional enhancements to consider when we continue working on it:
- Glow appears on initial state when steeping lab is opened
- We could possibly work non-determinism into this functionality

Issue #423, opened 2023-09-03
Enhance user-friendliness of Stepping Lab


S3865562:
### Proposed feature or solution

The Stepping Lab interface is currently quite minimalistic and could be more intuitive for users. Specifically, it lacks visual cues to indicate the state and progress of traversal.

Here are some proposed enhancements to improve the user-friendliness and functionality of the Stepping Lab:

- To prevent unnecessary clicks and confusion, disable buttons that would not result in a valid movement 
- Introduce a caret or another visual marker to pinpoint the exact segment of the input being processed during the trace tape traversal
- When the trace tape input is completely traversed or not accepted by the machine, display an alert or some form of visual cue to indicate this to the user
- When the Stepping Lab is opened, highlight the initial state to make it clear where the traversal begins

anioncat:
From user testing:

Stepping lab should tell us when its done stepping, e.g. blanking out the next button to not make it seem like it goes forever.

Stepping lab is also not clear on where in the input the pointer is. This is especially relevant in non-deterministic machines and related to https://github.com/automatarium/automatarium/issues/409 as the glow representing where the pointer is over laps with other glows and in some cases making it seem like the stepping has ended when it is just in a loop. Things that may help the problem:

- A display showing the input string as it progresses
- Changing the glow to a different colour to represent a state timeline has been accepted or rejected (green and red might cause confusion on green and red project themes)

Completing the merge in https://github.com/automatarium/automatarium/issues/405 may help with some clarity.

PR #429, opened 2023-09-15
Add enhancements/fixes to Stepping Lab

S3865562:
This update doesn't add all the enhancements originally aimed for -- disabling buttons when movement isn't possible relies on useState, which hasn't been behaving as expected in the SteppingLab component. Nevertheless, it does introduce several improvements and fixes a few bugs that emerged after the previous one was addressed.

Changes:

- Implemented Shift+2 keyboard shortcut for toggling Stepping Lab visibility.

- On mounting the Stepping Lab, the initial state is now highlighted.

- Integrated Stepping Lab support for TMs. Prior to this, utilising the Stepping Lab within a TM project led to an unhandled error causing the app to crash in dev mode.

- Removed stack visualiser when Testing Lab is inactive. While testing the Stepping Lab, it was found that the stack visualiser remained static throughout input string traversal. Given the Stepping Lab's inherent support for nondeterminism, omitting the stack visualiser outside of the Testing Lab seemed like a logical choice in order to avoid confusion.

Future Enhancements:
The initial issue (#423)  proposed a visual indicator for the character being processed during traversal and an alert or cue for unaccepted trace input. However, given the current support for nondeterminism in the Stepping Lab, these features don't actually seem viable. Since the Testing Lab already possesses these functionalities, their absence in the Stepping Lab isn't pressing.

Disabling buttons when movement isn't possible remains a priority.

### DIFF
--- a/frontend/src/components/Menubar/menus.ts
+++ b/frontend/src/components/Menubar/menus.ts
@@ -145,6 +145,10 @@ const menus: ContextItems = [
         action: 'TESTING_LAB'
       },
       {
+        label: 'Stepping lab',
+        action: 'STEPPING_LAB'
+      },
+      {
         label: 'File info',
         action: 'FILE_INFO'
       },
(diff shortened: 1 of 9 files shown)


JackGardnerRMIT (approved):
Looks fine to me

PR #433, opened 2023-09-22
Fixes left bounded TM tape

### LINKED ISSUE #361: Make Turing machine be infinite in both directions

Currently the Turing machine tape is only infinite to the right, unlike JFLAP which is infinite in both directions.

While a user should know the difference, it does make machines a bit more difficult since you need to add an extra character at the beginning so that you know when you are at the start again

### Alternatives

_No response_

### Additional information

_No response_

anioncat:
Closes #361.

The search could not transition if you expected a lambda at the front of the string, leading the TM tapes to be bounded on the left when they should be infinite.

image

Now:

image

Unsure how severely this would impact efficiency.


S3865562 (approved):
Thanks so much for doing this. If anyone decides to create the machine described in the upcoming TM tutorial video, it'll now work perfectly.

Issue #404, opened 2023-08-04
Trace Tape is Misaligned


giraugh:
### Description of the bug

When using the Testing Lab, the trace tape points to the wrong token. Seems like its always +2 from where it should be, including going off the end of the tape.

### To reproduce

Create simple FSA with two states and a transition
![image](https://github.com/automatarium/automatarium/assets/16392483/1b6c7c6e-1a84-4021-97f4-4493e941e068)

Enter a single "a" token into the testing lab and enable the trace tape
![image](https://github.com/automatarium/automatarium/assets/16392483/0773d3e9-70a4-43f2-9c2c-3bf8734f55a9)

The arrow on the floating trace tape is now misaligned
![image](https://github.com/automatarium/automatarium/assets/16392483/7b6d4f4b-24cf-4a10-be6d-3461198c337f)

### Expected behavior

Expected arrow to lineup with current token

### Additional information

![image](https://github.com/automatarium/automatarium/assets/16392483/803abfde-9713-487b-ac14-22638cbed8e7)

PR #434, opened 2023-09-23
Enhance multiple inputs per transition

### LINKED ISSUE #412: Multiple inputs per transition

It is very tedious to individually add every transition in a FSA.

Multiple transitions should be able to be added via comma separated values or similar.

### Alternatives

_No response_

### Additional information

(I know ranges exist somewhere so if this is already implemented then the issue is to document it for the users to know that they can use it.)

S3865562:
Closes #412 and #404 

- When a user enters a sequence of characters in the read field and presses enter, OR operators are inserted between each character ('abc' is now displayed as 'a | b | c')
- Enables multiple read inputs per transition for TMs, ensuring uniformity and consistency when working with different automata types
- Introduces a default OR operator which users can customise via the File Options menu
- Patches misaligned trace tape bubble

Screen Shot 2023-09-23 at 14 07 57

Screen Shot 2023-09-23 at 14 08 08

Currently working on a transition tutorial so people know about all of the unique things Automatarium can do with transitions.

### DIFF
--- a/frontend/src/components/InputDialogs/InputDialogs.tsx
+++ b/frontend/src/components/InputDialogs/InputDialogs.tsx
@@ -85,7 +85,6 @@ const InputDialogs = () => {
 
   const inputWriteRef = useRef()
   const inputDirectionRef = useRef()
-  const [read, setRead] = useState('')
   const [write, setWrite] = useState('')
   const [direction, setDirection] = useState<TMDirection | ''>('R')
   const editTransition = useProjectStore(s => s.editTransition)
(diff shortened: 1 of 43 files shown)


S3865562:
Unsure about including 'OR' as an option in the dropdown menu, definitely open to suggestions. 
Also going to prioritise increasing the margins between input fields for both TMs and PDAs, since with multiple inputs they can look especially cluttered.

anioncat:
Yeah I'm not sure on OR since a user might be using those characters or even $\lor$ since it looks a bit like v. Might be a bit confusing.

S3865562:
Have you experimented with it? At the moment, if the user enters 'v' but the ∨ operator is configured, the entered v is interpreted as ∨. So if they enter 'avb', it'll turn into 'a ∨ b'. I just figured if they want to use v in their input alphabet, they can just change the OR operator from the drop down. But if you think I should remove it, I'd be happy to.

I've used aliases for all of the symbols, another example is + and ＋.

anioncat:
I hadn't had a play with it yet but I now see that it's greyed out so it's pretty clear that its not part of the language. Is removing the or operator on the PDA pop/push string intended? Otherwise looks good to me

anioncat (approved):

PR #435, opened 2023-09-23
Transition dragging

### LINKED ISSUE #363: Drag transition to a new ending state

If you make a transition there is no way to move it to another ending state. If you want to move it then you need to delete it and recreate the transitions.

**Example**
image
There should be a circle or something at the end of the transition that I can drag so I can make the ending transition be $q_2$ instead of $q_1$

### Alternatives

_No response_

### Additional information

_No response_

anioncat:
Closes #363.

Adds smallish circles to the end of transitions that can be dragged to a different state.

image

Uses the `useTransitionCreation` hook to render a guide on where the selected transitions will be reallocated to. Only the selected transitions on the same edge will be moved.


S3865562 (approved):
So good!

ire4ever1190:
Nice! thats class

Tried it out but ran into this small bug where the transition didn't move as expected (i.e. didn't move like the ghost was showing)
Recording 2023-09-24 at 09 38 57

anioncat:
> Tried it out but ran into this small bug where the transition didn't move as expected (i.e. didn't move like the ghost was showing) 

Thanks for finding that. Should be fixed now.

Issue #272, opened 2023-03-07
Make direction for Turing machine transition maker be a dropdown


ire4ever1190:
The current UI just has arrows and doesn't communicate what the user should enter

A dropdown would present the valid values and make its purpose clearer

anioncat:
Closed with #436. Instead of drops they are radio button groups.

PR #436, opened 2023-09-24
Improve transition UX for PDAs and TMs

### LINKED ISSUE #406: Improve UX of transition creation for non-FSA automata

When creating transitions for PDAs and TMs, the transitions are more complicated and have more associated fields than for FSAs. We can improve the user experience by changing how these fields are represented.

Here's some ideas
- For PDAs, listing the fields horizontally rather than vertically
- For PDAs, adding margin between fields to visually seperate them
- And the ideas suggested in #272

S3865562:
Closes #406 and #272 

Changes:
- Increased margin between input fields for PDAs and TMs

Screen Shot 2023-09-25 at 00 01 41

- Implemented radio buttons (rather than a dropdown) for TM direction field

Screen Shot 2023-09-25 at 00 07 22

- Added radio buttons to TM direction field in InputTransitionsGroup editor

Screen Shot 2023-09-25 at 00 01 05


anioncat (approved):
Nice one. There's only the issue in the transition group editor where, when the radio buttons are focussed, the up and down arrow key behaviours conflicts with the default behaviour of moving the radio left/right, but I think that's too minor to block the PR (non-trivial as `e.preventDefault()` doesn't fix it and users have alternatives that work).

PR #441, opened 2023-09-28
Fix TM Tape Stretching Components

### LINKED ISSUE #440: UI elements hiden with long input string with turing machine
### Description of the bug

When I input a long string into the testing lab the left menu of the ui is hidden 

### To reproduce

here are some example strings:
`PXXXYI0000A0000B0000M0000R0000R0000R0000R0000R0000R0000R0000R0000`
`qwdqwdqwdqwdqwdqwqdwqwdwdwdwdqdqwwdqwdqwdqdwqwdwdqwdqqwdqwddwwwdqqdwwqwdqqwdwd`

### Expected behavior

I expect the ui to not move based on the string length 

### Additional information

Firefox 117 on a 1440p monitor

anioncat:
Closes #440.

Issue was caused by the whole tape taking space at all times and the animation being handled by a CSS transition (putting `margin-left: 1000px` on an element will get a similar effect).

This is a bandaid fix on the problem so that the TM tape for longer input strings is not obnoxious. A rewrite/redesign of the TM tape component might be needed to fix it completely. The fix will cull the size of the rendered tape so that it won't stretch the middle components, and maintain the animation (but janky) to keep up the appearance of the tape moving.

Alternative would be to make all the tape's parents not stretch but I couldn't for the life of me get it working (and you will still have the overflow problem).

### DIFF
--- a/frontend/src/components/TMTraceStepWindow/TMTraceStepWindow.tsx
+++ b/frontend/src/components/TMTraceStepWindow/TMTraceStepWindow.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import { useState, useEffect, useRef, useCallback } from 'react'
 
 import {
   Container,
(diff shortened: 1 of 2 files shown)


S3865562 (approved):
Fixes the issue 👍

PR #444, opened 2023-10-02
Thumbnails conform with app theme

### LINKED ISSUE #298: Update existing project thumbnails to match light/dark theme
Existing project thumbnails should update according to light/dark theme. This will be a matter of extending the dynamic colouring of the new project thumbnails to existing project thumbnails.

The mismatch between new project and existing project thumbnails to be addressed:
Screen Shot 2023-03-23 at 11 50 15 am

anioncat:
Closes #298. While I'm already working on thumbnails might as well push this through too.

- Thumbnails generate both light and dark modes in store so they can be swapped at will
- Thumbnails switch between a `-dark` and no (light) suffix and are generated on save, so thumbnails for old projects in dark mode will be broken (just make a change in a project to get a new one)
- This also applied to templates

#### Light
image
#### Dark
image

Ignore the broken ones, they got erased when I specified light with `-light`.

### DIFF
--- a/frontend/src/components/ProjectCard/projectCardStyle.ts
+++ b/frontend/src/components/ProjectCard/projectCardStyle.ts
@@ -30,7 +30,7 @@ export const CardImage = styled('div')<{$image?: boolean}>`
   align-items: center;
   justify-content: center;
   position: relative;
-  background: var(--grid-bg-light);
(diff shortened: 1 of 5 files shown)


S3865562 (approved):
Really improves dark mode, good job :)

PR #454, opened 2023-10-11
Exclusion inputs in transitions

### LINKED ISSUE #425: Exclusion (not) input in transitions

Making machines would be a lot easier if we could transition when a read value is not something. E.g. if we made a machine that doesn't change the tape and wanted to cycle back to the start after validating it.

### Alternatives

_No response_

### Additional information

_No response_

S3865562:
Closes #425 

This update introduces the capability to create exclusion inputs within transitions using a NOT operator (!). For instance, with an input such as !a, all symbols except 'a' are permissible. Users have three different ways to implement this feature:
1. Prefix a single character with !, e.g. !z
2. Apply ! before a character range, e.g. ![a-z]
3. Apply ! before expressions encapsulated in parentheses, where values are connected with an AND operator (&), e.g. !(a&c&d)

Screen Shot 2023-10-11 at 15 19 07

Changes:

- Input expressions prefixed with ! will be treated uniquely, permitting only symbols that are not present in the read

- Added new string manipulation functions that facilitate the formatting of exclusion inputs and the extraction of input symbols from a given string 

- Added tests to validate the formatting and extraction processes related to exclusion inputs

- Added tests to verify the functionality of exclusion transitions

-  Revised the info tab to appropriately display exclusion expressions

Screen Shot 2023-10-11 at 15 19 21

### DIFF
--- a/frontend/src/components/InputDialogs/InputDialogs.tsx
+++ b/frontend/src/components/InputDialogs/InputDialogs.tsx
@@ -6,7 +6,7 @@ import { useProjectStore, useViewStore } from '/src/stores'
 import useEvent from '/src/hooks/useEvent'
 import { locateTransition } from '/src/util/states'
 import { lerpPoints } from '/src/util/points'
-import { formatInput, splitCharsWithOr } from '/src/util/stringManipulations'
+import { formatInput, formatOutput } from '/src/util/stringManipulations'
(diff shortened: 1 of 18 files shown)


anioncat (approved):
Looks good. Going to be very useful

PR #455, opened 2023-10-16
Update keyboard shortcuts

S3865562:
Added the keyboard shortcut Shift+5 for toggling the Templates panel in preparation for recording a short tutorial video. Updated the guide to include the Stepping Lab and Templates shortcuts.

Noticed that 'Clear Selection' was listed in the Shortcut guide without a corresponding shortcut, so added 'Escape'.

Also added a Help menu option for accessing tutorial videos.

### DIFF
--- a/frontend/src/components/Menubar/menus.ts
+++ b/frontend/src/components/Menubar/menus.ts
@@ -188,6 +188,10 @@ const menus: ContextItems = [
         action: 'OPEN_DOCS'
       },
       {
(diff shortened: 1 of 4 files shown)


anioncat (approved):
Lgtm. There's not really a standard for clearing selection that I can think of, though I will throw out Esc and Ctrl+D (for deselect) as ones I've more commonly seen. (Not saying I haven't seen Ctrl+Shift+A either)

PR #456, opened 2023-10-17
Transition group editor fixes

### LINKED ISSUE #448: Transition Group UX Issues
### Description of the bug

1. Up and down keys do not function correctly when less than the full transition group is selected when bringing up the Transtion Group UI
2. Transition group editor for TMs do not have the new multi-read-input TM dialogs
3. Tabbing in the transition group should move to the next input dialog

### To reproduce

1.1. Create an edge with more than one transition
1.2. Delselect one of the transitions (shift + click)
1.3. Right click on edge to open transition group editor
1.4. Press down to see it not work properly

2.1. Right click on any edge and open the transition group editor
2.2. Attempt to enter more than one character in the read field for a new transition

3.1. Tabbing in the transition group tabs to the delete button at the end

### Expected behavior

1. Pressing up and down should cycle between all inputs sequentially
2. I should be able to enter more than one read input

### Additional information

1. The modal references are not correctly updated when the rest of the transitions are read in

anioncat:
Closes #448. TL;DR: Fixes all components of this issue.

> Up and down keys do not function correctly when less than the full transition group is selected when bringing up the Transtion Group UI
- This ended up being a project wide problem in that after we fire off the context menu we forgot the context of the event. The `<Component>EventData` information is forgotten when the context menu takes the position of the cursor to display it.
  - This can also be observed if you click on state 1, shift+click on state 2, right-click on state 2 and change the label; the label is changed on state 1. On the same essence righting clicking on edges in the same way has the same problem.
- To solve this, a context store was created to keep track of this. **TODO**: This needs to be propagated to the rest of the project and was out of scope of this one.
- Finally, moved references to be recreated as a side effect of the transition list changing enabled correct functionailty

> Transition group editor for TMs do not have the new multi-read-input TM dialogs
- The original editor directly updated the store on the go. This created a new problem in that formatting the input on the go means committing it into store.
- This created issues with state management with the inputs desyncing and with the exclusion inputs which no longer allowing the `!(x & y)` style inputs as it trunced to 1 character on change
- The input now operates on a temporary list of each input and commits to store on blur, or on submit in the case of the new input
  - This is consistent with the input dialog on the canvas

> Tabbing in the transition group should move to the next input dialog
- This is subjective, but delete and submit buttons have been removed from the tabbing index.

Side note:
- The input dialogs have also been refactored to their own functions above the giant switch statement into their respective sections. Refactoring them into their own components would be ideal, but passing references through props is causing problems.


S3865562 (approved):
Looks good :)

PR #458, opened 2023-10-18
Improve transition edge UX

### LINKED ISSUE #450: Increase hit box of transition handles

The handles to move transitions are too small and are hard to click, especially with a laptop trackpad. They should be larger.

### Alternatives

_No response_

### Additional information

Implement this similar to the way the edges have a path that is rendered and a path that is invisible with a larger width.

### LINKED ISSUE #451: Transition Edge UX

Clicking on an edge should always select all transitions on the edge. Right now if at least one transition is selected, clicking on the edge does nothing.

Edges with single transitions should be treated as the same as the transition. Double-clicking on the edge does nothing by default but allowing it to edit the one transition would improve the UX.

### Alternatives

On 2, we should still keep the group editor on there and grey out the single transition functions if a group is selected. Maybe we could combine the edge and transition contexts.

### Additional information

_No response_

anioncat:
Closes #451, closes #450.

Implements fixes in the above issues. Lumped these together as they both modify the edge.

- Transition dragging circles have a 10 radius (20 diametre) invisible circle as clicking error, the same thickness as the edge's invisible path. This is drawn on top of all the edge components so take priority.
- Clicking on an edge now always selects all transitions on the edge.
  - This gives a "select all transitions on edge" option so they can select all then deselect vs selecting each individually
  - A user right-clicking an edge can be assumed they want to act on that edge, but we can add a check for right-click if that assumption is incorrect
- Added extra context items to the edge context menu
  - Flip edge
  - Edit first transition (Double-clicking the edge already edits the first transition)
- Changed transition actions to use the context passing system so that the targeted item is the one being acted on
- Users pressing the `delete` key while in the context menu runs the global delete action, this keeps the menu open so added an event to close it


S3865562 (approved):
Looks good 👍

PR #459, opened 2023-10-18
Stop transition editing deleting transition on cancel

### LINKED ISSUE #449: Clicking on the canvas when editing transition deletes transition
### Description of the bug

This is quite annoying when you accidentally double click a transition.

### To reproduce

1. Create any transition
2. Double-click/right-click -> edit transition
3. Click on the canvas to cancel

### Expected behavior

Editing a transition should not delete a transition on cancel.

### Additional information

Interestingly you cannot undo/redo this change. Pressing undo will not do anything, but pressing redo again will undo the deleted transition.

anioncat:
Closes #449.

Edit transition receives an optional flag specifying whether a transtion being edited is a new transition or not. The input dialog receives that and does not go through with he deletion if the edit event is not for a new dialog. Defaults to `true`, which is the previous behaviour, so should not regress on any other part of the app.


S3865562 (approved):
Nice one.

PR #460, opened 2023-10-20
Add new tutorial videos to tutorial page

S3865562:
Added the following tutorial videos to the tutorial page:
- Creating projects & customising workspace (gave it the title 'Getting Started')
- Transitions
- Templates

Fixed issue where queries on links would break thumbnails.

Removed the advanced section for now. The 3 new videos don't have a lot in common (especially the Getting Started one), so it didn't feel right putting them all under the same section, and 6 videos on the main page isn't overwhelming. If future groups create new videos and want to add a section, they can use the [wiki](https://github.com/automatarium/automatarium/wiki/Tutorial-Videos) for guidance.


JackGardnerRMIT (approved):
Looks good to me

Issue #237, opened 2022-06-26
Automatic graph layout


GRA0007:
Following techniques outlined in [this paper](http://www.graphviz.org/documentation/TSE93.pdf), we should implement an automatic graph layout algorithm.

ire4ever1190:
One issue with the paper is needing to properly be able to control drawing transitions. Looking at JFLAP the degree tree layout algorithm seems to give pretty good results
**Before**: Using a messy graph where I just kept drawing transitions 
![before](https://user-images.githubusercontent.com/19339842/234847299-5a9de2ff-6eac-4c90-b9ba-7fdeb2b924fa.png)
**After**
![after](https://user-images.githubusercontent.com/19339842/234847180-4ebd0f4d-2ac0-4427-abe4-60c8e96bae06.png)

The algorithm seems pretty simple, ~200 lines* of Java in JFLAPs implementation (which also includes a hierarchical based algorithm along with degree based) 

*Excluding comments, also I might've miscounted

ire4ever1190:
We sadly ran out of time before being able to implement this.
Good luck group 4 🫡

anioncat:
I've begun progress in the `feat/auto-layout` branch but not happy enough with it to make a PR for the final sprint. It's been a headache so good luck future team 🫡.

PR #462, opened 2023-10-27
NFA -> DFA Basic Layouting

anioncat:
Not quite happy with auto layout (#237) yet but wanted to hook the NFA -> DFA converter to the basic ported layouter so the result stops stacking on top of each other. This will also merge the tests on the DAG creator.

Also fixes the warnings messages on NaN values for states when they are not created by the user.


S3865562 (approved):
Very nice 👍

Issue #405, opened 2023-08-04
Merge Testing and Stepping lab Functionality


giraugh:
### Proposed feature or solution

The functionality of the testing and stepping labs have a high degree of overlap. The stepping lab can have its unique features rolled into the testing lab rather than existing as a seperate tab.

PR #469, opened 2024-04-22
Completion of task 6

s3897720:
Task 6 of Automatarium v5 is completed (#405 resolved). I have opted to leave SteppingLab files for now, in case future developers need to see how the merge has occurred and step that back if major problems arise.

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
+++ b/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
@@ -263,16 +263,16 @@ const TestingLab = () => {
             disabled={traceIdx <= 0 || automataIsInvalid || noStepper
             }
             onClick={() => {
-              setTraceIdx(0);
-              handleStep('Reset');
+              setTraceIdx(0)
+              handleStep('Reset')
             }} />
 
           <Button icon={<ChevronLeft size={23} />}
             disabled={traceIdx <= 0 || automataIsInvalid || noStepper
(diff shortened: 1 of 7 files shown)


frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
```diff
@@ -52,6 +53,45 @@ const TestingLab = () => {
...
+    }
+  }
+
+  useEffect(() => {
+    if (stepper) {
+      handleStep('Reset')
+    }
+  }, [traceInput])
+
+  const noStepper = stepper === null && false
```
ire4ever1190, line 93: Apparently I wrote this line whoops
Its always false no? So can be safely removed

s3897720: Should be done now, thanks for picking that up. It's definitely always false.
Surely there's a reason why you've wrote that in the Stepping Lab... right? :)

ire4ever1190: 🤷‍♂️
Have a feeling it was debugging stuff I accidentally left in lol

s3782534 (approved):
Seems to mostly work, however the stepping doesn't move for the last transition if it would result in a success. I don't think it's a major issue as the execution trace shows which state the successful execution finished in.
Screenshot from 2024-05-08 11-35-01

Issue #359, opened 2023-04-30
State that is both initial and final considers no path to final


ire4ever1190:
### Description of the bug

Very minor bug that basically never would appear in real world. When there is a single state in the project and it is both the initial and final (with no transitions), automatarium thinks there is not path to the final state

### To reproduce

Build a machine like this
![image](https://user-images.githubusercontent.com/19339842/235337125-e9b23f46-270c-4427-bcdb-c20fe5013c25.png)

### Expected behavior

Should be accepted fine

### Additional information

This is caused by this line
https://github.com/automatarium/automatarium/blob/227b594220fa683c8fead1e692e34761ffb6eea6/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.js#L176

Since `closureWithPredicate` doesn't return the state it started with, it doesn't consider the initial node (Because a node can't transition to itself without a transition). Fix is easy, just adding the initial state into the returned list of nodes.

PR #477, opened 2024-05-06
Completion of issue #359

s3897720:
As title. Hardcoded a return of true in _pathToFinal_ for the edge case, also removed the path warning that was flagging for machines with 0 or 1 state(s), since a path cannot form in such number of states.

I'll leave the philosophical debate of "is it _really_ a path without transitions?"  to future groups.

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
+++ b/frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
@@ -175,8 +175,6 @@ const TestingLab = () => {
   const noInitialState = [null, undefined].includes(graph?.initialState) || !graph?.states.find(s => s.id === graph?.initialState)
   const noFinalState = !graph?.states.find(s => s.isFinal)
   const warnings = []
-  if (noInitialState) { warnings.push('There is no initial state') }
-  if (noFinalState) { warnings.push('There are no final states') }
 
   // Update disconnected warning
   const pathToFinal = useMemo(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/Sidepanel/Panels/TestingLab/TestingLab.tsx
```diff
@@ -180,10 +180,16 @@ const TestingLab = () => {
 
   // Update disconnected warning
   const pathToFinal = useMemo(() => {
-    const closure = closureWithPredicate(graph, graph.initialState, () => true)
-    return Array.from(closure).some(({ state }) => graph.states.find(s => s.id === state)?.isFinal)
```
ire4ever1190, line 184:
```ts
    // Since closure doesn't add initial
    return Array.from(closure).concat(graph.initalState).some(({ state }) => graph.states.find(s => s.id === state)?.isFinal)
```
Wouldn't this be enough?

ire4ever1190: Can then move the checking of final state existance down from 179 into something like this
```ts
if (noFinalState) { warnings.push('There are no final states') }
else if (!pathToFinal) { warnings.push('There is no path to final state') }
```
So that only one error is shown

s3897720: image
VSC is apparently not particularly happy with using _concat_ that way, but the idea is otherwise more sound than mine - I will look to restructure for this shortly.
You are right about the warnings though, that is a better way to do things (have fixed on my end, will look to push soon)

s3897720: Neat, that works. Thanks for that.

I'd also like to take a moment to appreciate the linter deciding to append it all to one line, from
`if (noFinalState) { warnings.push('There are no final states') }
else if (!pathToFinal) { warnings.push('There is no path to final state') }`
to
`if (noFinalState) { warnings.push('There are no final states') } else if (!pathToFinal) { warnings.push('There is no path to final state') }`

ire4ever1190: Linter can be a lil silly at times

ire4ever1190 (approved):
LGTM 🚀

PR #484, opened 2024-06-16
Removed test code from TMPause.

AntiToni:
I forgot to remove some test code in TMPause branch before it got pulled


s3782534 (approved):
Looks good

PR #490, opened 2024-07-26
Update share URL when opening export modal

### LINKED ISSUE #489: Need to reload page before beeing able to share the automaton
### Description of the bug

Hey!

While playing around with automatarium I noticed that sometimes I need to reload the page in order to use some functionalities properly. 
To give you an example: Lets say im creating a new project, I'm building an automaton and I than want to share my project by clicking on the button inside the menubar. If I'm doing all this without leaving the editor and without reloading the page in any other way it will not work as expected. Instead of showing the automaton I've created the link will only lead to an empty project. But if I reload the page before sharing my project everything works fine.
Its not that you always get an empty project when sharing without reloading, but It never contains the most recent changes made.

If you can't reproduce this behavior just hit me up and I will do a screen record:-)

At first I thought this might be a problem with auto saving the project, so I checked what project details are saved in my browsers local storage before reloading the page, and it looks fine.

It might still be connected to how and when the project is saved. I've also got similar problems with some functionalities I'm implementing myself, which aren't connected to the share-functionality

Are you aware of this problems? And do you know why it occurs?

Thanks for answering so quickly to my first question!

Greetings!

### To reproduce

_No response_

### Expected behavior

_No response_

### Additional information

_No response_

tdib:
Upon opening the share modal, the base 64 representing the project was only up to date as of the first load of the page.

This PR will update the share URL when the share button is pressed, meaning no more refreshing before sharing :)

Fixes #489

### DIFF
--- a/frontend/src/components/ShareUrl/ShareUrl.tsx
+++ b/frontend/src/components/ShareUrl/ShareUrl.tsx
@@ -14,15 +14,17 @@ const ShareUrl = () => {
 
   const [base64Project, setBase64Project] = useState('')
 
-  if (base64Project === '') {
-    encodeData(project).then(setBase64Project)
-  }
-
   const shareRawLink = `${window.location.origin}/share/raw/${base64Project}`
 
   const [exportUrlOpen, setExportUrlOpen] = useState(false)
   const [copySuccess, setCopySuccess] = useState(false)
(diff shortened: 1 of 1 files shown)


ire4ever1190 (approved):
LGTM 🚀

PR #493, opened 2024-07-26
Fix autosave status timing

### LINKED ISSUE #259: Reflect save state on save itself

tdib:
Autosaving status message now displays correctly according to when it is happening as opposed to having to wait for an interval timeout.

Not sure how "reacty" this approach is, but I've fixed the issue by having our `useAutosaveProject` hook return a value corresponding to whether or not the project is currently being saved, and passing this state to our menu bar.

Fixes #259

| Original | Updated |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4e1f2a27-98b5-4048-b483-d888e9aacacb"> | <video src="https://github.com/user-attachments/assets/bda3e9ba-fd45-4d63-ab32-ba3990f173be"> |

The previous approach also seems to have had an issue with indefinitely displaying the spinner when updating a project to have nothing, i.e. deleting everything in it. When any state or comment is added back to the project, the spinner goes away. This issue is now fixed.

| Indefinite load |
| --- |
| <video src="https://github.com/user-attachments/assets/89360086-0dee-4008-a0d5-27ca014cd804"> |

### DIFF
--- a/frontend/src/components/Menubar/Menubar.tsx
+++ b/frontend/src/components/Menubar/Menubar.tsx
@@ -71,6 +71,8 @@ const Menubar = ({ isSaving }: { isSaving: boolean }) => {
 
   const projectName = useProjectStore(s => s.project?.meta?.name)
   const setProjectName = useProjectStore(s => s.setName)
+  const lastChangeDate = useProjectStore(s => s.lastChangeDate)
+  const lastSaveDate = useProjectStore(s => s.lastSaveDate)
   const setLastSaveDate = useProjectStore(s => s.setLastSaveDate)
   const upsertProject = useProjectsStore(s => s.upsertProject)
 
@@ -95,10 +97,10 @@ const Menubar = ({ isSaving }: { isSaving: boolean }) => {
   }
 
   useEvent('beforeunload', e => {
-    if (!isSaving) return
+    if (lastSaveDate > lastChangeDate) return
     e.preventDefault()
     return 'Your project isn\'t saved yet, are you sure you want to leave?'
-  }, [isSaving], { options: { capture: true }, target: window })
+  }, [lastSaveDate, lastChangeDate], { options: { capture: true }, target: window })
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


ire4ever1190:
Think it might be unintended side effect with the removal of the last change check but now empty projects are saved.
e.g. a user clicks to create a new FSA project, immediately regrets it and clicks out, now an empty project is saved that they'll probably delete

tdib:
ah good catch. is there a nice way to prevent this from happening? I originally considered removing this `isSaving` conditional and saving regardless (why wouldn't we save when we're leaving) - would there be any concerns with removing that check to instead include a condition to check for empty project? e.g.

```
if (projectNotEmpty) {
   saveProject()
}
```

ire4ever1190:
Can't forsee any issues? Think that should be fine

tdib:
That should do it - I also noticed the unsaved project warning was broken so I've gone ahead and fixed that too

Just to confirm the behaviour is that we'll never save a completely empty project and any time you exit the project from that state it will disappear?

tdib:
Agh so if you have a project with states, and you delete them then immediately go to the projects page, the project doesn't save and it just keeps your machine as is, but the thumbnail shows it empty

Will have to sort this issue out tmr or sumn

ire4ever1190:
in my mind the behaviour is more like if the user hasn't touched it then don't save it. So guess could like check if lastChangeTime is not set and then not save, otherwise do? Think that'd be least surprising behaviour

tdib:
> if the user hasn't touched it then don't save it
So this behaviour is already working, but the thing I just described kinda leaves the UI and project in a weird state bc the project itself doesn't save, but the thumbnail makes it look like it has

I'm thinking it might actually make sense to just delete a project if there's nothing on it - if they're leaving it, the history gets wiped anyway so when they go back it's almost exactly fresh (aside from the name). Similar behaviour to say text documents where if you open it and close it you won't get a prompt to save, and if you type and then remove everything you still won't get a prompt

ire4ever1190:
ah just saw that comment lol
But yeah if that makes sense then do it that way 👌 😎

tdib:
I think the behaviour you described seems easy enough to implement (assuming there's a way to tell the user hasn't interacted with the project), but I'm kind of stumped on how to have the project delete if it's empty/get the same behaviour as if it was never interacted with

tdib:
I thought that one would be super simple after discovering that we had a `deleteProject` function but that caused a bunch of issues with our image generation hook in some cases so I've had to add some null logic around the place

The one case that is still ever so slightly broken is if you
1. Open a project, add some states, and save it
2. Go back into said project, and delete all states - this will not autosave
3. Open a new tab, and you can see the empty project thumbnail, but clicking into it will still contain your states

ire4ever1190 (approved):

PR #501, opened 2024-07-28
Expand read symbols for DFA conversion 

### LINKED ISSUE #445: NFA to DFA conversion did not produce DFA
### Description of the bug

Hitting Convert NFA -> DFA on this machine will produce the same machine which is not an NFA (`q0` has more than one transition for `b`).

image

I don't have the original machine anymore since I only realised this after saving and exiting the editor, so I don't have the base NFA I converted with in the first place.

Possibly caused by `b` and `a b` being counted as different inputs in the NFA to DFA algorithm.

### To reproduce

_No response_

### Expected behavior

I get an DFA.

### Additional information

Probably not useful but here's the original jumble after converting once. The above is after I pruned anything not necessary.

image

ire4ever1190:
Closes #445

Core of the issue was how the DFA conversion handled multiple symbols for single transition (Hint, it did not handle it). Makes me wonder if we should change the `read` property in transitions to be an array so type checking would pick this up in future 🤔

Take this example NFA
image

Jflap convert it into this

image

While current code converts it to this (Notice how there are two transitions for `a`)
image

After fix it gets converted into this
image


packages/simulation/src/convert.ts
```diff
@@ -9,6 +9,14 @@ const makeStateLabel = (states: number[]): string => {
   return states.length === 0 ? 'Trap' : states.sort((a, b) => a - b).join(', ')
 }
 
+/**
+ * Returns all read symbols for a transition
+ */
+const readSymbols = (t: FSAAutomataTransition): string[] => {
+  // Either split all the character, or return a lambda transition
+  return t.read ? t.read.split('') : ['']
```
tdib, line 17: Do you need to chuck `expandReadSymbols(t.read)` here for read strings like `[a-z]`?

packages/simulation/tests/convertNFAtoDFA.test.ts
```diff
@@ -11,6 +11,8 @@ import convertSingleTrapState from './graphs/convertSingleTrapState.json'
 import convertSingleTrapStateDFA from './graphs/convertSingleTrapStateDFA.json'
 import infiniteLoop from './graphs/infiniteLoop.json'
 import infiniteLoopDFA from './graphs/infiniteLoopDFA.json'
+import issue445Expected from './graphs/issue445DFA.json'
+import issue445Input from './graphs/issue445NFA.json'
```
tdib, line 15: Weird naming but go off king 🤪 (I'd name it something more relevant to the contents of the machine, someone in the future isn't really gonna know what issue 445 is)

tdib (commented):
Just the one consideration about using `expandReadSymbols`, not sure if it's necessary in this context (like maybe the symbols are already expanded) but yeah


> Makes me wonder if we should change the read property in transitions to be an array so type checking would pick this up in future 🤔

Dunno if this would be a trivial change bc for cases like `[a-z]`, we store the actual string `[a-z]` in the read and just expand that at simulation time. I actually considered how we'd even change this but when I realised we'd have to go the other way (i.e. from separate reads back to the display form `[a-z]`) I was like hells naw

packages/simulation/src/convert.ts
ire4ever1190: Oh shoot forgot about that.
Think that function might handle everything actually

packages/simulation/tests/convertNFAtoDFA.test.ts
ire4ever1190: Smh people should know every issue number in their head before even knowing what automatarium is 😤

But ye will update that name, and chuck in a test case for the [a-z] thing

ire4ever1190:
I forgot we support negation :sob: will need to think a bit more about how this conversion process will work

ire4ever1190:
Used that existing expand function, much nicer.

So kinda have an idea in my head on how to handle negations, but will do that in a separate PR since it will require quite a bit more work. Want to add a proper parsing of transitions instead of the adhoc method that is done now, once its all structured then adding support for negations should be a lot easier

tdib:
So with your current approach is it just support for NFA -> DFA conversion (minus negation)?

ire4ever1190:
Correct 
Probs won't be able to get around to the two issues I created for a while so will be good to least have this bug fixed for CT students

PR #531, opened 2025-04-11
Update disabled button styling for module remove button

### LINKED ISSUE #520: Last question "Remove" button colour
When a module only has one question, the "Remove" button should use the disabled colour rather than the red colour to indicate that it will not perform any action. This goes for any other buttons with no-behaviour

Image

tdib:
Closes #520 

This PR adds disabled button styling for the module remove question button to indicate that it is not clickable

| Original | Updated |
| --- | --- |
| image | image |

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/Modules/modulesStyle.ts
+++ b/frontend/src/components/Sidepanel/Panels/Modules/modulesStyle.ts
@@ -68,6 +68,10 @@ export const RemoveButton = styled('button')`
   align-items: center;
   justify-content: center;
   transition: background-color 0.3s;
+
+  &:disabled {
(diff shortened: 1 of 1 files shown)


ire4ever1190:
Failed to take screenshot cause my cursor gets hidden
But the cursor still thinks it's clickable, think it's another bug tho

tdib:
> cursor still thinks it's clickable

Oh valid, tbh I thought that was the behaviour we had elsewhere, lemme see if it's an easy fix

tdib:
It's fixed but is that the proper way? Feels hacky

ire4ever1190 (approved):
Seems fine
Brief read in the docs and didn't see anything else

PR #532, opened 2025-04-11
feat: add confirmation dialog when deleting module questions (#521)

### LINKED ISSUE #521: Remove question confirmation
In a module, when a user presses "Remove", there is no confirmation for the action that is performed. Since we don't want a user to lose their work, we should add confirmation before the deletion occurs

https://github.com/user-attachments/assets/8a0d1b49-e505-4eff-8e82-0a16ec209a20

hanoon-07:
Closes #521 

-Description
*Added confirmation modal when removing a question from a module
*Prevents accidental deletion of questions and their associated content
*When there is only a single question, there is no option given to remove

Changes made in: 
frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
frontend/src/components/Sidepanel/Panels/Modules/modalStyles.ts (added a small margin top for aesthetic purposes)

-Issue Context
In a module, when a user presses "Remove", there is no confirmation for the action that is performed. Since we don't want a user to lose their work, we should add confirmation before the deletion occurs

https://github.com/user-attachments/assets/72f018f9-aa29-46c1-8853-65fe3f8ce7f3

🙌 Additional Notes
Happy to contribute once more in this project :) . My second ever Open source contribution.

### DIFF
--- a/frontend/src/components/Modal/modalStyle.ts
+++ b/frontend/src/components/Modal/modalStyle.ts
@@ -63,6 +63,7 @@ export const Description = styled('span')`
   display: block;
   padding-inline: 1rem;
   opacity: .5;
+  margin-top: 0.5rem;
 `
 
 export const Buttons = styled('div')`
--- a/frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
+++ b/frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
@@ -39,6 +39,10 @@ const Modules = () => {
 
   // Modal-related state management
   const [isModalOpen, setIsModalOpen] = useState(false) // Controls modal visibility
+  
+  // Delete confirmation modal state
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false)
+  const [projectToDelete, setProjectToDelete] = useState<ModuleProject | null>(null)
 
   // Load module values into local state when module changes
   useEffect(() => {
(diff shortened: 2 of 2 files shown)


tdib:
Nice stuff, can you make sure to run `yarn lint` (or `yarn lint --fix`) and fix any of the issues? Looks like there's a few issues with trailing spaces

hanoon-07:
sure, on it!.

frontend/src/components/Modal/modalStyle.ts
```diff
@@ -63,6 +63,7 @@ export const Description = styled('span')`
   display: block;
   padding-inline: 1rem;
   opacity: .5;
+  margin-top: 0.5rem;
```
tdib, line 66: Not sure we should add extra margin - this modal is used in quite a few places, can you revert this?

hanoon-07: yes i do understand, but the spacing provides better look wherever the modal is used if you notice.

frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
```diff
@@ -236,19 +255,22 @@ const Modules = () => {
                     <td onClick={() => handleOpenQuestion(q)}>{`Question ${index + 1}`}</td>
                     <td>
                       <EditButton onClick={() => handleEditQuestion(q)}>Edit</EditButton>
-                      <RemoveButton
+
+                      {currentModule.projects.length > 1 && <RemoveButton
```
tdib, line 259: We still want to show the remove button even if there's only one - I've recently added some styling to disable the button

Reason for this is so the buttons and their locations are familiar, so when the user adds more questions, they already know there will be a remove button
```suggestion
                      <RemoveButton
```

frontend/src/components/Modal/modalStyle.ts
hanoon-07: should i still revert it? Please let me know.

hanoon-07:
Hey ive made changes keeping the remove button even if there is one single question. About the extra margin, i think it makes the modal look better. If im wrong please let me know, open to suggestions.

tdib:
> About the extra margin, i think it makes the modal look better

tbh I was looking at the export image modal which imo looks better with less spacing, but other modals like delete template look better with more spacing. Happy to keep it as is since most modals are closer to delete template


Export image
| Original | Updated |
| --- | --- |
| image | image | 

Delete template
| Original | Updated |
| --- | --- |
| image | image |

frontend/src/components/Modal/modalStyle.ts
tdib: We can keep it :) https://github.com/automatarium/automatarium/pull/532#issuecomment-2798269363

tdib (approved):
Good stuff!

hanoon-07:
> Good stuff!

thanks!

ire4ever1190 (approved):

PR #546, opened 2025-08-26
Set up github pages for pr preview

### LINKED ISSUE #534: Deploy using github pages

Currently the site is deployed with vercel which gives the neat feature of auto deployments for PRs.
The downside is that since it uses the free tier I can't add other users onto the group.

Should look into managing the deployments via github actions since that it all runs off the same system and can be managed by students in the automatarium group

### Alternatives

_No response_

### Additional information

Need to make sure the DNS records are updated to point to the new deployment

alyuhhh:


.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set PR preview env variable
+        shell: bash
+        run: echo 'PR_PREVIEW=pr-preview/pr-${{ github.event.number }}' >> .env
```
ire4ever1190:
```suggestion
        run: echo 'PR_PREVIEW=/automatarium/pr-preview/pr-${{ github.event.number }}' >> .env
```

.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
...
+  deploy-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set PR preview env variable
+        shell: bash
+        run: echo 'PR_PREVIEW=/automatarium/pr-preview/pr-${{ github.event.number }}' >> .env
+
+      - uses: ./.github/actions/setup
```
ire4ever1190, line 24: oh lol, think the checkout in this step is overwring the .env file

.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set PR preview env variable
+        shell: bash
+        run: echo 'PR_PREVIEW=/automatarium/pr-preview/pr-${{ github.event.number }}' >> frontend/.env
```
ire4ever1190: Should this maybe be something generic like `SUB_PATH`?

frontend/src/index.tsx
```diff
@@ -84,11 +84,13 @@ const App = () => {
   </>
 }
 
+const basename: string | undefined = process.env.PR_PREVIEW ? `/${process.env.PR_PREVIEW}` : undefined
```
ire4ever1190:
```suggestion
// Should have a leading `/`
const basename = process.env.PR_PREVIEW ?? "/"
```

.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
+name: Deploy PR previews
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+
+concurrency: preview-${{ github.ref }}
```
cloudyapple4: The syntax for using concurrency might have since changed from the example given in `rossjrw/pr-preview-action`
Reading the [github actions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency), I can only see examples that look like this:
```suggestion
concurrency: 
  group: preview-${{ github.ref }}
  cancel-in-progress: true
```

cloudyapple4 (changes requested):
Overall, previews are looking good. 

Do we also want to deploy a production version as well? I know this branch/PR is only concerned with previews, but the linked issue sounds like we're gonna replace all deployments with GH pages.

ire4ever1190:
+1 for production version also. But in another PR please, I don't want to keep approving vercel deployments haha

cloudyapple4 (approved):

alyuhhh:
Aim is to change prod to GH Pages. This was to test it out

PR #551, opened 2025-08-27
Refactor/replace aom with json

### LINKED ISSUE #510: Replace `.aom` and `.ao` formats with `.json`
The original choice behind `.json` as a file extension was to be able to open it in most text editors and have a known format. `.aom` and `.ao` are not known formats, and hence cannot be opened as easily.

Update `.aom` and `.ao` export methods to use `.json` extensions - these formats can be distinguished using a field within the json such as "type" within the metadata. If there is a need for more distinction, the module files may be exported as `.module.json`, which would still adhere to using a json format.

s4013303:
Closes issue #510 

## Changes
- replaced .ao file with .json in frontend/public/locales/bg/common.json
- replaced .aom file with .json in frontend/public/locales/bg/common.json
- replaced .ao file with .json in frontend/public/locales/en/common.json
- replaced .aom file with .json in frontend/public/locales/en/common.json
- removed .ao file in frontend/src/components/ImportDialog/ImportDialog.tsx
- replaced .ao file with .json in frontend/src/components/ImportDialog/ImportDialog.tsx
- replaced .aom file with .json in frontend/src/components/ImportModuleDialog/ImportModuleDialog.tsx
- removed .ao file in frontend/src/hooks/useActions.ts
- replaced .ao file with .json in frontend/src/hooks/useActions.ts
- replaced .aom file with .json in frontend/src/hooks/useActions.ts

## Testing
- Created an FSA, exported as `.json`, and successfully imported it back.


s4013303:
I believe it added the pull request Refactor: remove duplicate enum from Modules.tsx and import from ProjectTypes #550 to this as well, both work

frontend/src/hooks/useActions.ts
```diff
@@ -153,7 +153,7 @@ const useActions = (registerHotkeys = false) => {
         const file = new Blob([JSON.stringify(project, null, 2)], { type: 'application/json' })
         a.href = URL.createObjectURL(file)
         // File extension explicitly added to allow for file names with dots
-        a.download = project.meta.name.replace(/[#%&{}\\<>*?/$!'":@+`|=]/g, '') + '.ao'
+        a.download = project.meta.name.replace(/[#%&{}\\<>*?/$!'":@+`|=]/g, '') + '.json'
         a.click()
```
ire4ever1190, line 157: Do you mind throwing this into a function while you are here? Think I forgot to create an issue over it.

Or just throw it in an issue

frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
```diff
@@ -8,6 +8,8 @@ import { exportModuleFile } from '/src/hooks/useActions'
 import { dispatchCustomEvent } from '/src/util/events'
 import { Plus } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
+// Import enum from ProjectTypes
```
tdib: Comment probs not necessary?

tdib (commented):
Thanks for doing this :) Will hold off on approving until @ire4ever1190's comment is addressed, otherwise looks good!

s4013303:
I threw @ire4ever1190's comment in an issue and removed the unnecessary comment, should all be resolved now

tdib:
On a second look, it seems like this doesn't actually support .aom? It imports it as a regular project and the contents vanish, so this would actually break a few things

It looks like the module format is a little different to the standard file, so perhaps we can reuse the `projectType` property from FSA/PDA/TM and that's how we can differ the projects

ire4ever1190 (approved):

s4013303:
I thought that was the point, to remove .ao and .aom and to instead replace it with .json? would you like me to add .json alongside .ao and .aom so it all works together?

tdib:
> I thought that was the point, to remove .ao and .aom and to instead replace it with .json? would you like me to add .json alongside .ao and .aom so it all works together?

Sorry nah you've got the right understanding, but I mean taking a `.aom` file, renaming to `.json`, and trying to import doesn't work

s4013303:
I see, so what did you need instead of this implementation?

tdib:
> I see, so what did you need instead of this implementation?

Just realised the modules need to be imported in a different place, i was trying to import it as a project. Implementation works as expected, apologies

I'll raise a separate issue for this because it's a little misleading that it allowed me to import a module as a project, and it looked like it didn't work properly

frontend/src/hooks/useActions.ts
tdib: #569

tdib (approved):

s4013303:
Sounds good, appreciate the review and feedback

PR #553, opened 2025-08-28
Action to upload dist on release

alyuhhh:
Action to easily retrieve dist folder to send to RMIT library


.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+          node-version: 'lts/*'
+
+      - name: Enabling Corepack
+        run: corepack enable
+
+      - name: Installing Latest Yarn
+        run: yarn set version stable && yarn install
+
+      - name: Install Yarn Dependencies
+        run: yarn
```
ire4ever1190: Could you look into moving this into like a setup.yaml [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action), just so it's not repeated everywhere

ire4ever1190 (changes requested):

ire4ever1190 (approved):

PR #555, opened 2025-09-03
fix(module): rename Question → Section in module_panel labels

### LINKED ISSUE #523: Reword "Question" in modules
Within modules, the term "Question" may not be the most fitting. Its usage is likely by a tutor/lecturer explaining a concept as opposed to asking something. This could be adjusted to fit more broadly, such as "Chapter" or "Section"

Image

Image

Jnemhara:
- Replaced all instances of "Question" with "Section" in module_panel translations
- Updated JSON strings to reflect new naming convention
- Verified UI now shows "Section 1, Section 2..." instead of "Question 1..."

Before
Screenshot 2025-09-04 053235

After
Screenshot 2025-09-04 053130


frontend/public/locales/en/common.json
ire4ever1190, line 1: Might be good idea to update the Bulgarian files so they stay in sync

tdib:
Seems like this closes #523? If so, can you update your description so the issue closes when this gets merged?

PR #556, opened 2025-09-03
fix(questions): make Cancel use secondary style and lock hover (#524)

### LINKED ISSUE #524: Cancel button on question creation does not follow secondary button styling
The cancel button on the questions panel does not follow the secondary button styling convention. Currently, it is an outlined button that, when hovered, turns into a secondary button styling. This button should match others by using the secondary styling without a hover effect

https://github.com/user-attachments/assets/86e24a83-dc51-4263-b0c1-eeda389fa5ba

Desired styling:
image

Jnemhara:
Cancel button in the Questions panel was styled as outline and morphed on hover.

Updated Button usage to pass secondary prop.

Adjusted ButtonStyleClass to ensure $secondary stays solid (no hover change).

Before

Screenshot 2025-09-04 044105

After
Screenshot 2025-09-04 043951


frontend/src/components/Button/buttonStyle.tsx
```diff
@@ -0,0 +1,56 @@
+import { css } from 'goober'
+
+type WrapperProps = {
+  $icon: boolean
+  $surface: boolean
+  $secondary: boolean
+}
+
+/**
+ * Creates a wapper element. This is done to support thr `as` prop with typescript
```
tdib, line 10:
```suggestion
 * Creates a wapper element. This is done to support the `as` prop with typescript
```

frontend/src/components/Button/buttonStyle.tsx
```diff
@@ -0,0 +1,56 @@
...
+      : 'var(--primary) !important',   
+
+    // Hover styles
+    '&:hover': $secondary
+      ? { backgroundColor: 'var(--toolbar)' }
+      : $surface
+      ? { backgroundColor: 'var(--surface)' }
+      : { filter: 'brightness(0.96)' },
+
+    '&:active': { filter: 'brightness(0.94)' },
```
tdib, line 45: Is this changing the styling of all secondary buttons across the site? It feels a bit strange using an arbitrary value here instead of a central variable in case we want to replicate similar behaviour elsewhere

tdib (commented):
image
image


Not sure what's happened but it seems to look different on my end?

Issue #557, opened 2025-09-04
Save button in module creation is lowercase


tdib:
### Description of the bug

<img width="345" height="322" alt="Image" src="https://github.com/user-attachments/assets/3c558438-7fb6-4380-820a-adeb3605b56e" />

The save button is labelled with a lowercase s, which is inconsistent with other buttons across the site.

### To reproduce

1. Create a project
2. Click on modules (right side)
3. Click on modularise this project
4. Enter any name and create
5. Click on edit
6. The save button is lowercase

https://github.com/user-attachments/assets/816825b9-6e48-44da-bd38-1c0f5ad5cc10

### Expected behavior

The s should be uppercase :)

Elham-EN:
Hi! I'd like to work on this issue. I'm new to open source and this seems like a great first contribution. 

I plan to:
- Locate the save button in module creation that's displaying in lowercase
- Update it to proper capitalization 
- Follow the project's git conventions with a fix/ branch
- Submit a PR to the dev branch

Could you assign this issue to me? Thanks!

PR #558, opened 2025-09-04
fix: capitalize save button text in module creation

Elham-EN:
**Fixes issue #557**

This PR fixes the capitalization of the save button in the module creation interface.

**Changes:**
- Updated translation key "save" from lowercase "save" to "Save" in public/locales/en/common.json
- This affects save buttons throughout the module creation interface
- Improves UI consistency with proper button capitalization

**Testing:**
- ✅ Verified save buttons now display "Save" with proper capitalization
- ✅ No functional changes, only visual improvement
- ✅ Linting passes (`yarn lint`)

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,22 @@
+name: "Setup environment"
+description: "Setup node"
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v5
+    - name: Set up Node.js Latest LTS
+      uses: actions/setup-node@v4
+      with:
+        node-version: 'lts/*'
+
+    - name: Enabling Corepack
+      shell: bash
+      run: corepack enable
+
+    - name: Installing Latest Yarn
+      shell: bash
+      run: yarn set version stable
+
+    - name: Install Yarn Dependencies
+      shell: bash
+      run: yarn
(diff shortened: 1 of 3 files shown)


ire4ever1190:
I think you might've ran the wrong lint config? Thats seems to be a lot more formatting changes than what I'd expect

ire4ever1190:
Think you might've ran the wrong lint config? Thats a lot more changes than I would've expected

Elham-EN:
Hi! I wanted to clarify what happened with my latest update:

**Original issue**: The commit included both the translation file fix AND unintended formatting changes to the React component file.

**What I fixed**: 
- ✅ Removed the React component file changes (that was just auto-formatting, not needed for the fix)
- ✅ Kept only the translation file change from "save" to "Save"

**Current status**: The PR now shows only 1 file changed (the translation JSON), though it appears an auto-formatter also cleaned up the JSON structure throughout the file. The actual functional change is still just the "save" → "Save" update.

Let me know if you'd prefer a more minimal diff that only touches the specific line, or if the JSON formatting cleanup is actually helpful to keep.

ire4ever1190:
Sorry, do you mind just keeping the change to just the save button?
Though this does raise a good point that the translation files are formatted differently (Created #560 to track that)

PR #563, opened 2025-09-14
Replace numbered pagination with 'x of y' QuestionPager

### LINKED ISSUE #522: Pagination double up behaviour
When creating multiple questions in modules, we can only ever see 3 questions at once, so you can either go to the question right before, or right after where you currently are. This just so happens to be the same behaviour as the left and right arrows.

This should be reconsidered as to not double up button behaviour. I've attached an Automatarium image when we have 5 questions (it shows 3), and a proposed design.

Essentially we should only show the current question, and the number of questions, allowing the user to 1) navigate to the previous and next questions, and 2) navigate to any question of their choice, by typing into the input (which indicates the current question)

Current design:
image

Proposed design:
Image

Jnemhara:
Replaced the old numbered pagination UI in ModuleWindow with a new `QuestionPager` component.

- Removed duplicated numbered button logic
- Added new QuestionPager component in `src/components`
- Uses "x of y" display with arrows and direct input
- Styled with CSS Modules for scoped styles
- Updated both Editor and Example ModuleWindow pages

Simplifies navigation, avoids duplicated button behavior, and matches proposed design from issue #522 

Before:
Screenshot 2025-09-15 024142

After:
Screenshot 2025-09-15 024235
Screenshot 2025-09-15 024402
Screenshot 2025-09-15 024438


ire4ever1190:
Hi looks good, though you might want to move this into goober elements? (Just for consistency with rest of project)

PR #565, opened 2025-09-17
Integrate questions panel into module window

Jnemhara:
### Changes
- Removed left `ModuleWindow` question editor from Editor layout
- Updated sidebar (`Modules.tsx`) to handle add/edit/remove questions
- Added inline placeholder "Edit module instructions here"

Simplifies UX by consolidating all question editing into the right sidebar.

Before
Before

After
FUll
Edit question


frontend/src/components/Sidepanel/Panels/Modules/Modules.tsx
```diff
@@ -1,286 +1,236 @@
...
-                      >
-                        {t('module_panel.remove')}
-                      </RemoveButton>
-                    </td>
-                  </tr>
-                ))}
-              </tbody>
-            </Table>
-            <Button icon={<Plus/>} onClick={handleAddQuestionClick}>{t('module_panel.add_question')}</Button>
+                    {currentModule.questions[id] || "Edit module instructions here"}
```
ire4ever1190: rhs should be a translation

frontend/src/components/Sidepanel/Panels/Modules/moduleWindowStyling.ts
```diff
@@ -0,0 +1,147 @@
...
+`
+
+export const TitleWrapper = styled('div')`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 1rem;
+`
+
+export const Title = styled('h2')`
```
ire4ever1190: some of these seem very generic. Like there is already a `Title` element defined elsewhere https://github.com/automatarium/automatarium/blob/660e3af621efc99095f14995f30684e5cf015a4f/frontend/src/pages/Tutorials/tutorialsStyle.ts#L10

Would some existing components work instead of defining them again?

frontend/src/stores/useModuleStore.ts
```diff
@@ -45,39 +40,33 @@ export const createNewModuleProject = (projectType: ProjectType = DEFAULT_PROJEC
     statePrefix: DEFAULT_STATE_PREFIX,
     orOperator: DEFAULT_OR_OPERATOR,
     acceptanceCriteria: DEFAULT_ACCEPTANCE_CRITERIA,
-    // All lab projects are set to pink to distinguish the difference between a lab and a normal project
```
ire4ever1190, line 48: Some of these comments are helpful

.yarn/install-state.gz
ire4ever1190, line 1: Throw this in the .gitignore while you're at it

frontend/src/stores/useModuleStore.ts
```diff
@@ -41,6 +41,8 @@ export const createNewModuleProject = (
     orOperator: DEFAULT_OR_OPERATOR,
     acceptanceCriteria: DEFAULT_ACCEPTANCE_CRITERIA,
     color: 'pink'
+        // All lab projects are set to pink to distinguish the difference between a lab and a normal project
```
ire4ever1190, line 44:
```suggestion
    // All lab projects are set to pink to distinguish the difference between a lab and a normal project
   color: 'pink'
```

package-lock.json
ire4ever1190, line 1: Yarn is the package manager used, not npm

ire4ever1190:
The UI looks to be jutting out weirdly for me
image

@tdib might have a better opinion, but maybe the delete should be better distinguised? Think we have a secondary button somewhere
image

tdib:
> @tdib might have a better opinion, but maybe the delete should be better distinguised? Think we have a secondary button somewhere

Yeah agreed, I reckon we can use the `--error` colour (red) here which should give it a bit of distinction/hierarchy

image

PR #575, opened 2025-09-24
Fix thumbnails using wrong theme with `system` theme

ire4ever1190:
The CSS before was `prefer-color-scheme: dark` when it should be `(prefers-color-scheme: dark)` which lead to it never matching and always using light theme when user had system theme selected.

Also fixed the debouncing code which wasn't properly clearing the last timeout, small optimisation

**Before**
image

**After**
image

### DIFF
--- a/frontend/src/components/Sidepanel/Panels/Templates/Templates.tsx
+++ b/frontend/src/components/Sidepanel/Panels/Templates/Templates.tsx
@@ -35,7 +35,7 @@ const Templates = () => {
   const thumbs = useThumbnailStore(s => s.thumbnails)
 
   const theme = usePreferencesStore(state => state.getTheme())
-    const getThumbTheme = useCallback((id: string) => {
+  const getThumbTheme = useCallback((id: string) => {
       const thumbTheme = theme === 'dark' ? '-dark' : ''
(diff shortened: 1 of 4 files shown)


frontend/src/components/Sidepanel/Panels/Templates/Templates.tsx
```diff
@@ -33,13 +33,11 @@ const Templates = () => {
...
   const thumbs = useThumbnailStore(s => s.thumbnails)
-  const theme = usePreferencesStore(s => s.preferences).theme
 
-  const getThumbTheme = useCallback((id: string) => {
-    const thumbTheme = theme === 'system'
-      ? window.matchMedia && window.matchMedia('prefer-color-scheme: dark').matches ? '-dark' : ''
-      : theme === 'dark' ? '-dark' : ''
-    return `tmp${id}${thumbTheme}`
+  const theme = usePreferencesStore(state => state.getTheme())
+    const getThumbTheme = useCallback((id: string) => {
```
ire4ever1190:
```suggestion
  const getThumbTheme = useCallback((id: string) => {
```

tdib (approved):

PR #581, opened 2025-10-01
Deploy main and dev to github pages

alyuhhh:


ire4ever1190:
Can you check against the warning here https://github.com/rossjrw/pr-preview-action?tab=readme-ov-file#ensure-your-main-deployment-is-compatible 
Just checking that the gh-pages thing doesn't overwrite the previews

cloudyapple4:
Looks like dev and main will deploy to the same url. Do we want to keep them separate?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,26 @@
+name: Deploy to GitHub Pages
```
cloudyapple4, line 1:
```suggestion
name: Deploy Production
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,26 @@
...
+    - uses: ./.github/actions/setup
+
+    - name: Build project
+      run: yarn build
+
+    - name: Deploy to GitHub Pages
+      uses: JamesIves/github-pages-deploy-action@v4
+      with:
+        folder: ./frontend/dist/
+        clean-exclude: pr-preview/
```
cloudyapple4:
```suggestion
        clean-exclude: pr-preview/
        force: false
```

cloudyapple4 (changes requested):
How do you guys want to proceed with deploying the dev preview? 
`JamesIves/github-pages-deploy-action` has `target-folder` to push the build to a subfolder. With the right set up, we could have dev deploy to `/automatarium/preview`. We'd just need to make sure it plays nice with the other deployments.

alyuhhh:
yea was thinking of creating another workflow that's specific to dev. not sure if that's the best way tho like if there's a way to keep in one

cloudyapple4:
> yea was thinking of creating another workflow that's specific to dev. not sure if that's the best way tho like if there's a way to keep in one

That sounds fine to me. I'd go with that if you haven't already

ire4ever1190 (approved):

cloudyapple4 (approved):

PR #582, opened 2025-10-02
Offline Support via PWA

### LINKED ISSUE #544: Add offline support via PWA

Enable users to download Automatarium to use offline via Progressive Web App (PWA) standards.

### Alternatives

_No response_

### Additional information

_No response_

cloudyapple4:
Closes #544 

This PR adds features to support the offline use of Automatarium:

#### `manifest.json` 
Provides basic metadata for the browser to install the web app. Installing doesn't enable offline use. It just integrates the web app with the device so it can be accessed and opened like a native app. 

image

#### `service-worker.js` 
Enables offline use of the app. It automatically precaches all necessary files to use the app offline.
- Once cached, resources that are cached will always be served from the cache. (Updates are supported, as detailed in the next section)
- A temporary toast is shown at the bottom of the page when all resources have been cached to notify a user that the app is available offline. This is also shown on every subsequent loading of the app, just in case it was missed the first time.

    Screenshot 2025-10-15 at 2 48 57 pm

- Service workers and their caches are limited to their scope. Deployments on different subfolders of the same domain will each have their own service worker and cache, preventing clashes on GH Pages deployments.
- The app currently takes about 6.3 MB of storage.

#### `build-sw.js` 
Injects the service worker with a list files found in `dist/` so it knows what to precache. It is run during the build process after parcel builds `dist/`.
- Gifs, which are hosted externally, have been added as additional entries to the cache list.
- Fonts have been moved to being self-hosted to make precaching easier.

## Updates
Updates are handled through this process:
1. Changes to files cause their hash to change. 
2. The service worker's cache manifest (which includes hashes) will also update when rebuilt.
3. When loading the app, the browser checks if the service worker has changed. If so, it installs the new one.
4. The new service worker takes over, updating the cache in the background.
5. The cache is now updated. And updated content will be served on the next load.
6. A persistent toast is shown in the bottom corner to notify the user to reload for updated content.

Screenshot 2025-10-15 at 2 52 42 pm

## Tutorials Page
When offline, the tutorials page is replaced with a warning instead of displaying broken youtube links.

Screenshot 2025-10-15 at 2 53 34 pm


frontend/build-sw.ts
ire4ever1190, line 1: Do you mind making the file be typescript? Just for consistency

frontend/package.json
```diff
@@ -2,12 +2,15 @@
...
   "private": true,
   "scripts": {
     "start": "yarn prebuild && parcel public/index.html --no-cache",
     "copy:locales": "cp -r public/locales dist",
-    "prebuild": "rm -rf dist && mkdir dist && yarn copy:locales",
-    "build": "yarn prebuild && parcel build public/index.html --public-url ./",
+    "prebuild": "rm -rf dist && mkdir dist && yarn copy:locales && cp public/ping.txt dist/",
+    "build": "yarn prebuild && parcel build public/index.html --public-url ./ && yarn postbuild",
+    "postbuild": "node build-sw.js",
+    "serve": "serve dist -l 3000",
```
ire4ever1190, line 13: Q: What is serve for?

frontend/public/icon.png
ire4ever1190, line 1: You can use the existing SVG iirc, just set sizes to "any" in the manifest

frontend/public/service-worker.js
ire4ever1190, line 1: Also here, though don't know if it'll integrate well with the build system so not biggest of problems

frontend/package.json
cloudyapple4: I've been using it for local testing instead of the dev server because I can't run the postbuild script with hot reloading.

frontend/public/icon.png
cloudyapple4: I did try using the SVG originally, but this error was showing up in devtools (3rd line).

Screenshot 2025-10-15 at 5 13 18 pm

Despite the error, the icon appeared fine when installing. Maybe it's just a problem with devtools? I ended up using a png in an attempt to resolve it. I couldn't find a way to fix it while using the svg.

With that being said, it should be fine to switch back to the SVG. Lemme know if you'd prefer that!

ire4ever1190: Strange, failed to load feels like a different probably (I'm assuming you had correct type set also)?

Would prefer reusing the existing svg if possible

ire4ever1190 (approved):

PR #583, opened 2025-10-07
Fix favicon theming

cloudyapple4:
Fixed the regex so the favicon updates to match the accent color. The problem was an extra space after the comma.
I also removed a pair of parentheses because they were redundant.


ire4ever1190:
Cool! Didn't realise this was a feature

ire4ever1190 (approved):

PR #588, opened 2025-10-23
Automatically email builds

alyuhhh:
Added a workflow that would automatically send an email with the zipped build every time a push is made into main so RMIT Library deployment stays in sync

Need testing in dev before changing it to main tho


ire4ever1190:
So will I get spammed with an email that I need to then forward on?

alyuhhh:
> So will I get spammed with an email that I need to then forward on?

No, my bad. I didn't need to set it up like that for testing

ire4ever1190 (approved):

